PR #5, opened 2016-08-15
show client name as well as campaign name

steppenwells:
the campaign name on its own can be ambiguous.

Small PR but wanted to get into the frontend code and get a feel for propTypes etc.


clloyd:
I was going to do this in a slightly different way... basically a separate component that allows you to view the client information but 👍 and can modify later

steppenwells:
This is more of a composite title per campaign rather than an attempt to show more client information.

the compostite title is used so that if we have multiple TBC campaign names (we do) they can be identified by client. Same argument works for not just showing client name.

PR #7, opened 2016-08-23
Edit Campaign

clloyd:
This adds the ability to edit name/value, and save. Will add more fields as needed

screen shot 2016-08-23 at 13 41 51


steppenwells:
Looks good :+1:

I'm going to merge this so I can fold it into the analytics stuff before adding utils etc.

Quick question though - the `CampaignPerformanceMetrics` component, is this a placeholder for the analytics display component in my branch, or is this earmarked for something else?

PR #10, opened 2016-08-30
Polishing colour palette of graphs

Calanthe:
This is little bit better but still not perfect. Some of the graphs overlaps on each other and their colours also mix. It's hard to predict how it will look like on different data sets but imo for now it is ok. Colours are more soften and less vivid.

Before:
screen shot 2016-08-30 at 15 34 52

After:
screen shot 2016-08-30 at 15 26 03


clloyd:
👍  Looks better to me!

PR #12, opened 2016-08-31
Campaign value validation

Calanthe:
It irritated me that after writing any random characters in the `campaign value` field I didn't see any errors and the input was still empty. That happened because we didn't validate that field correctly. This PR aims to fix that. 

I want to change the `error` state in the parent component only (`CampaignInformationEdit`) so I had to move `enableEditing()` and `disableEditing()` from it's child component `EditableText` to the aforementioned `CampaignInformationEdit`. I am not sure if this is the best approach.

I also don't know why do we have `defaultProps` static value in `EditableText` module. I could't find any usage of it. Should I remove it?

Screen:
screen shot 2016-08-31 at 14 31 56


clloyd:
Unfortunately, I wouldn't move the state up a level, it is specific to the individual field being edited so should be part of the chid component.

If you wanted the subcomponent to show an error, then you could pass the error as a prop. Then the parent can make the decision to show it or not.

Calanthe:
@clloyd I moved the validation logic to the parent component (`CampaignInformationEdit.js`) and only send error statues to the child component (`EditableText.js`).

Please review the code again.

clloyd (approved):

PR #15, opened 2016-09-02
Editable campaign status

Calanthe:
This PR introduces editable status value (as a dropdown) in the campaign details view.
I also added the 'last modified' info.

screen shot 2016-09-02 at 16 09 13

### DIFF
--- a/public/components/CampaignInformationEdit/CampaignInformationEdit.js
+++ b/public/components/CampaignInformationEdit/CampaignInformationEdit.js
@@ -1,9 +1,17 @@
 import React, { PropTypes } from 'react';
 import EditableText from '../utils/EditableText';
+import EditableDropdown from '../utils/EditableDropdown';
 
 import {formatMillisecondDate} from '../../util/dateFormatter';
 
 class CampaignInformationEdit extends React.Component {
+
+  stateValues = { //Not sure if this is the best place for it. Maybe we should have dedicated module with those values and import it here, in sidebar and filters?
+    live: 'Live',
+    prospect: 'Prospect',
+    production: 'Production',
+    dead: 'Dead'
+  }
 
   state = {
     isCampaignDirty: false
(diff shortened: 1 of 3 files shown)


public/components/CampaignInformationEdit/CampaignInformationEdit.js
```diff
@@ -1,10 +1,18 @@
 import React, { PropTypes } from 'react';
 import EditableText from '../utils/EditableText';
+import EditableDropdown from '../utils/EditableDropdown';
 
 import {formatMillisecondDate} from '../../util/dateFormatter';
 
 class CampaignInformationEdit extends React.Component {
 
+  stateValues = { //Not sure if this is the best place for it. Maybe we should have dedicated module with those values and import it here, in sidebar and filters?
```
clloyd: I would move these to a module in /public/constants

public/components/utils/EditableDropdown.js
```diff
@@ -0,0 +1,69 @@
...
+    this.setState({
+      editable: false
+    });
+  }
+
+  prepareOptions = () => {
+    const values = this.props.values;
+    let options = [];
+
+    for (let prop in values) {
```
clloyd: Can we make this more functional, I think a filter and map or a reduce would work well.

public/components/CampaignInformationEdit/CampaignInformationEdit.js
```diff
@@ -54,6 +62,7 @@ class CampaignInformationEdit extends React.Component {
   }
 
   render () {
+      console.log(campaignStatuses);
```
clloyd: console.log

public/components/utils/EditableDropdown.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  prepareOptions = () => {
+    const values = this.props.values;
+    let options = [];
+
+    values.map((prop, index) =>
+      options.push(<option value={prop.value} key={index}>{prop.name}</option>)
+    );
+
```
clloyd, line 35: So this could be written as 

```
return values.map((prop, index) =>
      return <option value={prop.value} key={prop.value}>{prop.name}</option>
 );
```

clloyd:
small comments but looks great!

PR #16, opened 2016-09-05
Client Editing

clloyd:


clloyd:
cc @steppenwells

app/controllers/ClientApi.scala
```diff
@@ -0,0 +1,73 @@
...
+        ClientRepository.putClient(client)
+        Ok(Json.toJson(client))
+      }
+    }
+  }
+
+
+  def bootstrapData() = APIAuthAction { req =>
+
+    val user = loggedInUser(req.user)
```
steppenwells, line 36: I don't think this is being used.

steppenwells:
Looks ok, one minor comment.

👍

PR #18, opened 2016-09-08
disable editing when clicking outside input

Reettaphant:
- disable editing editing when clicking outside the input field instead of when clicking a disable button.


Reettaphant:
@clloyd

clloyd:
👍

PR #19, opened 2016-09-09
remove remove buttom from editable number

Reettaphant:
Remove the remove when editing numbers as well.


Reettaphant:
@clloyd I just noticed the editable number, the disable button should probably be removed from there as well.

clloyd:
👍

PR #21, opened 2016-09-20
Add notes

Reettaphant:
Adds the ability to view, create and update notes associated with a campaign


Reettaphant:
@clloyd

app/controllers/CampaignApi.scala
```diff
@@ -35,6 +35,66 @@ class CampaignApi(override val wsClient: WSClient) extends Controller with Panda
...
+  def getCampaignNotes(id: String) =  APIAuthAction { req =>
+    Ok(Json.toJson(CampaignNotesRepository.getNotesForCampaign(id)))
+  }
+
+  def addCampaignNote(id: String) = APIAuthAction { req =>
+
+    val content = (req.body.asJson.get \ "content").as[String]
+
+    if (content.isEmpty())
+      InternalServerError("Cannot add a note with no content")
```
steppenwells: BadRequest rather than InternalServerError, it's the requester that'd be at fault here.

app/controllers/CampaignApi.scala
```diff
@@ -35,6 +35,66 @@ class CampaignApi(override val wsClient: WSClient) extends Controller with Panda
...
+    }
+  }
+
+  def updateCampaignNote(id: String) = APIAuthAction { req =>
+
+    val dateCreated = (req.body.asJson.get \ "created").as[DateTime]
+    CampaignNotesRepository.getNote(id, dateCreated) match {
+      case None => BadRequest(s"Could not find note with id $id and create time $dateCreated")
+      case Some(note) => {
+        println("got the note from the repo ", note)
```
steppenwells: remove println or convert to log message.

app/controllers/CampaignApi.scala
```diff
@@ -35,6 +35,66 @@ class CampaignApi(override val wsClient: WSClient) extends Controller with Panda
...
+      CampaignNotesRepository.putNote(newNote)
+      Ok(Json.toJson(newNote))
+    }
+  }
+
+  def updateCampaignNote(id: String) = APIAuthAction { req =>
+
+    val dateCreated = (req.body.asJson.get \ "created").as[DateTime]
+    CampaignNotesRepository.getNote(id, dateCreated) match {
+      case None => BadRequest(s"Could not find note with id $id and create time $dateCreated")
```
steppenwells: Hmm interesting. Normally things not being found would be a 404 NotFound, however you're using the request json to pass the date portion of the composite key so technically BadRequest is also fine.

I would however consider moving the created date param into the URL (even if it gets duplicated in the request body) and use this to lookup the note. At the risk of channeling Seb I think this would be more RESTful as it has a fully addressable note in the url. If we go the URL route the error here would be a conventional 404

app/controllers/CampaignApi.scala
```diff
@@ -35,6 +35,66 @@ class CampaignApi(override val wsClient: WSClient) extends Controller with Panda
...
+      case Some(note) => {
+        println("got the note from the repo ", note)
+
+        val lastModified = DateTime.now()
+        val modifiedBy = User(req.user.firstName, req.user.lastName, req.user.email)
+        val content = (req.body.asJson.get \ "content").as[String]
+
+        val updatedNote = note.copy(
+          campaignId = note.campaignId,
```
steppenwells: When using a case class' copy function you only need to add the params that are changing, so `campaignId = note.campaignId` is redundant. This copy can be just:

```
val updatedNote = note.copy(
  lastModified = lastModified,
  lastModifiedBy = modifiedBy,
  content = content
)
```

conf/routes
```diff
@@ -19,7 +19,10 @@ GET /api/campaigns/bootstrap    controllers.CampaignApi.bootstrapData
 GET /api/campaigns              controllers.CampaignApi.getAllCampaigns
 GET /api/campaigns/:id          controllers.CampaignApi.getCampaign(id: String)
 PUT /api/campaigns/:id          controllers.CampaignApi.updateCampaign(id: String)
-GET /api/campaigns/:id/analytics          controllers.CampaignApi.getCampaignAnalytics(id: String)
+GET /api/campaigns/:id/analytics       controllers.CampaignApi.getCampaignAnalytics(id: String)
+GET /api/campaigns/:id/notes    controllers.CampaignApi.getCampaignNotes(id: String)
+POST /api/campaigns/:id/notes   controllers.CampaignApi.addCampaignNote(id: String)
+PUT /api/campaigns/:id/notes    controllers.CampaignApi.updateCampaignNote(id: String)
```
steppenwells: Related to discussion in the API code this url would be

`PUT /api/campaigns/:id/note/:date`

if we moved the date addressing to the url.

Since this is a PUT that only updates one note I would push for putting the date in the url. The GET on the same path gets all notes and the GET and PUT should really be symettric

public/components/Campaign/Campaign.js
```diff
@@ -1,5 +1,6 @@
 import React, { PropTypes } from 'react';
 import CampaignEdit from '../CampaignInformationEdit/CampaignEdit';
+import CampaignInformationEdit from '../CampaignInformationEdit/CampaignInformationEdit';
```
steppenwells: Unused?

public/components/CampaignNotes/CampaignNotes.js
```diff
@@ -0,0 +1,61 @@
...
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.campaign.id !== this.props.campaign.id) {
+      this.props.campaignNoteActions.getCampaignNotes(nextProps.campaign.id);
+    }
+
+  }
+
+  saveNote = () => {
+      this.props.campaignNotesAddActions.createNote(this.props.id, { content: this.state.message });
+      window.removeEventListener('click', this.disableAdding, true);
```
steppenwells: i'm sure there's a more react'y way to do this. @clloyd is probably better placed than me to advise.

steppenwells (changes requested):
I've looked at the server side and have some comments around the url scheme, RESTfulness etc, but otherwise ok.

I had a quick look at the client side react stuff but I think @clloyd is better placed to advise around that.

I would suggest maybe creating an editable TextArea component similar to the editable Text on we already have.

conf/routes
Reettaphant: I'm not sure what you mean, I've now changed the put route as you suggested and left the get route as it is since it is getting all the notes in the campaign.

public/components/CampaignNotes/CampaignNotes.js
clloyd: Yeah the removal of the EventListener should really be happening in the same component as it's being added (So in the future if this component is replaced then the new one wouldn't need to also do this). It will mean that you can't directly pass down the createNote function, but I think t's worth it

clloyd (changes requested):
Looks fine with that one client side change

public/components/CampaignNotes/CampaignNotes.js
Reettaphant: This saveNote function is actually not used anywhere, I forgot to remove when i was moving stuff around. I've now deleted and the actual removal and adding of the event listener are both happening in the campaignNotesAdd component

clloyd (approved):

PR #25, opened 2016-10-12
Sw analytics improvements

steppenwells:
The analytics shown should now always correspond to the campaign being looked at.

Locked the sidebar (Shareef don't like it)


steppenwells:
oh and cc @clloyd who worked with me on these

clloyd (approved):

PR #28, opened 2016-10-13
cache GA analytics results till the end of the day.

steppenwells:
Adds a simple cache that will store analytics results 'til the end of the day.


kelvin-chappell (approved):
An admin route to clear the cache manually might also be useful, although I suppose you could just redeploy.

PR #30, opened 2016-10-18
Add traffic-driver summary box to individual campaign page

kelvin-chappell:
image


steppenwells (commented):
Looks good, 👍 etc.

Obviously you need to fix the conflict with build.sbt but go ahead and merge once that's fixed

PR #31, opened 2016-10-19
Sw capi refresh

steppenwells:
Adds a refresh button to refetch / update content from CAPI.

screen shot 2016-10-19 at 12 08 21


app/model/command/ImportCampaignFromCAPICommand.scala
```diff
@@ -62,6 +46,61 @@ case class ImportCampaignFromCAPICommand(
...
+    val ctaAtoms = apiContent.flatMap(_.atoms.flatMap(_.cta)).flatten
+
+    val updatedCampaign = campaign.copy(
+      startDate = startDate.map{cdt => new DateTime(cdt.dateTime).withTimeAtStartOfDay()},
+      status = if(startDate.isDefined) "live" else campaign.status,
+      callToActions = ctaAtoms.map(ctaAtom => CallToAction(Some(ctaAtom.id))).distinct,
+      campaignLogo = deriveTagLogo(hostedTag)
+    )
+
+    CampaignRepository.putCampaign(updatedCampaign)
```
kelvin-chappell, line 83: Does this do an update or insert?  Assume it won't do an insert if ID exists but difficult to tell.

kelvin-chappell (approved):
Looks good and I assume that the semantics of 'put' are like map.put.

app/model/command/ImportCampaignFromCAPICommand.scala
steppenwells: It will do an insert or update depending on if the content record is already there.

PR #35, opened 2016-10-24
Aggregate traffic driver types in summary box

kelvin-chappell:
This change aggregates different types of traffic driver at the top line, and links through to the individual contributing DFP line items.

(Some imagination required as it's quite longwinded to get impressions on the test DFP site.)
Before:
image

After:
image


steppenwells (commented):
👍

PR #36, opened 2016-11-01
Show daily traffic driver stats

kelvin-chappell:
Now adds a component to the end of the campaign analytics showing stats for traffic driver impressions, clicks and CTRs by type of driver, linked to from the stats summary box:

image

/cc @steppenwells

### DIFF
--- a/public/components/CampaignAnalytics/Analytics/CampaignTrafficDriverStatsChart.js
+++ b/public/components/CampaignAnalytics/Analytics/CampaignTrafficDriverStatsChart.js
@@ -1,0 +1,172 @@
+import React, {PropTypes} from "react";
+import {
+  LineChart,
+  Line,
+  XAxis,
+  YAxis,
+  CartesianGrid,
+  Tooltip,
+  Legend,
+  ResponsiveContainer,
+  ReferenceLine,
+  Brush
+} from "recharts";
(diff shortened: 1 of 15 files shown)


public/components/CampaignAnalytics/Analytics/CampaignTrafficDriverStatsChart.js
```diff
@@ -0,0 +1,172 @@
...
+
+    const anchorName = group.groupName.toLowerCase().split(' ').join('-') + '-drivers';
+
+    const shortDateFormat = (date) => {
+      return new Date(date).toLocaleDateString('en-GB', {day: '2-digit', month: '2-digit'})
+    };
+
+    const longerDateFormat = (date) => {
+      return new Date(date).toLocaleDateString('en-GB', {weekday: 'short', day: '2-digit', month: 'short'})
+    };
```
steppenwells: There is a date format helper js file that is used to process dates elsewhere in the js layer (see dateFormatter.js). Might be worth moving these date ops into that file or checking if the existing formatter functions do the job already

steppenwells (approved):
Small comment re js date formats but otherwise all good

👍 

Once I've finished farting about with analytics, caches etc it might be worth checking if we want to use that for the DFP stuff but for now all's good

PR #38, opened 2016-11-02
Fix typos

lps88:
Receive has been spelled wrong in lots of places... bit of an OCD PR!!


steppenwells (approved):
Excellent. I thought about doing this the other day but couldn't face it.

👍

PR #39, opened 2016-11-02
Sw paid content prep

steppenwells:
This is the first of 2 PRs for adding paid content support. This adds a new field to the campaign model to indicate the type of campaign - hosted or paidContent.

The field is modelled as optional in this PR and there is a migration to populate all existing campaigns as hosted. In the second PR this field is no longer optional and it should only be applied once this has been deployed and the migration run.


app/controllers/Migration.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  def addCampaignType() = APIAuthAction { req =>
+    val campaigns = CampaignRepository.getAllCampaigns
+
+    campaigns foreach { c =>
+      val cWithType = c.copy(`type` = Some("hosted"))
+      CampaignRepository.putCampaign(cWithType)
+    }
+
+    Ok(s"added type hosted to ${campaigns.length} camapaigns")
```
kelvin-chappell, line 17: campaigns

kelvin-chappell (approved):
👍

PR #40, opened 2016-11-02
Sw paid content

steppenwells:
This is the second PR for paid content support, the first being https://github.com/guardian/campaign-central/pull/39

This provides paid content support and also reworks how we cache analytics data.

paid content support:

* paid content tag types can be selected in CAPI import
* all campaigns can be filtered by type as well as status
* reworked the sidebar to allow different types of filters to be applied
* removed display of CTA information for paid content campaigns on the info screen

analytics cache:

* added a second mechanism to fetch analytics for paid content as paid content does not all sit under a nice exclusive url prefix - this is unfortunately pretty slow when there is a lot of content
* store analytics results in a dynamo table so that only one of the node in the cluster has to do the expensive fetch - the dynamo table is defined in cloudformation and will need to be added before we roll this out
* add a management screen to allow the cache to be viewed and refreshed - this is a bit rough and ready (i.e it doesn't give any feedback when you perform operations and doesn't refresh automatically) but it'll only be used by power users (the dev team principally)
* serve stale for analytics data and fetch in the background so there should always be something to show


app/repositories/AnalyticsDataCache.scala
```diff
@@ -0,0 +1,91 @@
...
+import com.amazonaws.services.dynamodbv2.document.Item
+import com.amazonaws.services.dynamodbv2.document.spec.ScanSpec
+import play.api.Logger
+import play.api.libs.json.{Format, Json}
+import services.Dynamo
+
+import scala.util.control.NonFatal
+import scala.collection.JavaConversions._
+
+abstract sealed trait CacheResult[+A] extends Product
```
kelvin-chappell, line 13: Aren't all traits abstract?

kelvin-chappell (approved):
Looks fine.  Will make things work much more smoothly with multiple servers.

PR #42, opened 2016-11-07
Sw analytics summary

steppenwells:
Stores the headline campaign stats as a separate objects and as a rollup report for use when rendering campaign summaries.

### DIFF
--- a/app/repositories/AnalyticsDataCache.scala
+++ b/app/repositories/AnalyticsDataCache.scala
@@ -70,11 +70,29 @@ object AnalyticsDataCache {
     Dynamo.analyticsDataCacheTable.putItem(entry.toItem)
   }
 
+  def putCampaignSummary(campaignId: String, data: CampaignSummary, expires: Option[Long]): Unit = {
+    val entry = AnalyticsDataCacheEntry(campaignId, "CampaignSummary", Json.toJson(data).toString(), expires, System.currentTimeMillis())
+    Dynamo.analyticsDataCacheTable.putItem(entry.toItem)
+  }
+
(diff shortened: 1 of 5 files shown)


app/repositories/AnalyticsDataCache.scala
```diff
@@ -10,7 +10,15 @@ import services.Dynamo
 import scala.util.control.NonFatal
 import scala.collection.JavaConversions._
 
-abstract sealed trait CacheResult[+A] extends Product
+abstract sealed trait CacheResult[+A] extends Product {
```
kelvin-chappell: `Option` is abstract because it isn't a trait: http://www.scala-lang.org/api/current/scala/Option.html

app/repositories/AnalyticsDataCache.scala
```diff
@@ -60,16 +68,26 @@ object AnalyticsDataCacheEntry {
...
     Dynamo.analyticsDataCacheTable.deleteItem("key", key, "dataType", dataType)
   }
 
-
   def putCampaignDailyCountsReport(campaignId: String, data: CampaignDailyCountsReport, expires: Option[Long]): Unit = {
     val entry = AnalyticsDataCacheEntry(campaignId, "CampaignDailyCountsReport", Json.toJson(data).toString(), expires, System.currentTimeMillis())
     Dynamo.analyticsDataCacheTable.putItem(entry.toItem)
   }
 
+  def putCampaignSummary(campaignId: String, data: CampaignSummary, expires: Option[Long]): Unit = {
```
kelvin-chappell: Perhaps it should be clearer that `expires` is a time instant rather than an interval of time.

kelvin-chappell (approved):
👍

PR #43, opened 2016-11-07
compress analytics data when storing to dynamo

steppenwells:
Once the paid content stuff was in the wild it turned out some analytics data was too big to fit in a dynamo entry (Aviva etc). This introduces gzip compression and stores the analytics as a binary blob.

On read the compressed data field is checked, if it exists then use this if not assume the entry is just storing a JSON string and use that.

for comparison here are the after and before records for explore Canada side by side in atom:

screen shot 2016-11-07 at 17 25 39

I couldn't reduce the font enough to fit all of the json on the screen - an improvement I'm sure you'll agree

### DIFF
--- a/app/util/Compression.scala
+++ b/app/util/Compression.scala
@@ -1,0 +1,28 @@
+package util
+
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream}
+import java.util.zip.{GZIPInputStream, GZIPOutputStream}
+
+import org.apache.commons.io.IOUtils
+
+
+object Compression {
+
+  def compress(data: String): Array[Byte] = {
+    val arrOutputStream = new ByteArrayOutputStream()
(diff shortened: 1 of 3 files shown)


app/util/Compression.scala
```diff
@@ -0,0 +1,28 @@
...
+
+object Compression {
+
+  def compress(data: String): Array[Byte] = {
+    val arrOutputStream = new ByteArrayOutputStream()
+    val zipOutputStream = new GZIPOutputStream(arrOutputStream)
+    zipOutputStream.write(data.getBytes)
+    zipOutputStream.close()
+    val bytes = (arrOutputStream.toByteArray)
+    arrOutputStream.close()
```
kelvin-chappell, line 17: Isn't the underlying stream closed by `close()` in line 15?

kelvin-chappell (approved):
Nice 👍

PR #44, opened 2016-11-08
Campaigns in rows

Calanthe:
The campaign list should be a set of rows instead of grid of blocks:

Before:
screen shot 2016-11-08 at 10 26 00

After:
screen shot 2016-11-08 at 10 23 32

There will be a few more columns with some basic analytics info added later on.

### DIFF
--- a/public/components/CampaignList/CampaignListItem.js
+++ b/public/components/CampaignList/CampaignListItem.js
@@ -38,32 +38,14 @@ class CampaignListItem extends React.Component {
     }
 
     return (
-      <Link className="campaign-list__item" to={"/campaign/" + this.props.campaign.id}>
-          {image}
-        <div className="campaign-list__item__info">
-          <span className="campaign-list__item__info-name">{this.props.campaign.name}</span>
-          <span className="campaign-list__item__info-other">
-            <span className="campaign-list__item__info-other--info">Type: </span>
-            <span className="campaign-list__item__info-other--status">{this.props.campaign.type} </span>
-          </span>
-          <span className="campaign-list__item__info-other">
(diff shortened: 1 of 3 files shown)


public/components/CampaignList/CampaignListItem.js
```diff
@@ -38,32 +42,14 @@ class CampaignListItem extends React.Component {
...
-                <span className="campaign-list__item__info-other--info">Finish date: </span>
-                <span className="campaign-list__item__info-other--value">{endDate}{daysLeft}</span>
-          </span>
-        </div>
-      </Link>
+      <tr className="campaign-list__row" onClick={this.redirectToCampaign}>
+        <td className="campaign-list__item">{this.props.campaign.name}<br/>{image}</td>
+        <td className="campaign-list__item">{this.props.campaign.type}</td>
+        <td className="campaign-list__item">{this.props.campaign.status}</td>
+        <td className="campaign-list__item">{this.props.campaign.actualValue}</td>
```
kelvin-chappell, line 49: What is the `actualValue` of a campaign?

kelvin-chappell (approved):
👍

public/components/CampaignList/CampaignListItem.js
Calanthe: I don't know. @steppenwells ?

steppenwells: This derives from the pipeline spreadsheet where campaigns had actual and nominal values. Now we're selling hosted properly the actual value should be non-zero but we're modelling both to allow for the kind of creative accounting that business people love.

PR #45, opened 2016-11-08
Small UI fixes on campaigns list

Calanthe:
Small UI fixe on campaigns list:
- logos' maximum width is set to 100px


lps88:
Cool 👍

PR #46, opened 2016-11-08
UI enhancements

lps88:
- Add hover states to buttons
 - Make filter menu items into links (so clicking on one from another section brings you back to campaigns list)
 - update react-router version so that highlighting works properly in the menu

### DIFF
--- a/public/components/Sidebar/Sidebar.js
+++ b/public/components/Sidebar/Sidebar.js
@@ -19,7 +19,7 @@ class Sidebar extends React.Component {
     }
 
     return(
-      <div className={className} onClick={() => this.setCampaignStateFilter(stateValue)}>{displayName}</div>
+      <Link to="/campaigns" className={className} onClick={() => {this.setCampaignStateFilter(stateValue)}}>{displayName}</Link>
     )
   };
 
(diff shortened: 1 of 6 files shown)


public/components/Sidebar/Sidebar.js
```diff
@@ -19,7 +19,7 @@ class Sidebar extends React.Component {
     }
 
     return(
-      <div className={className} onClick={() => this.setCampaignStateFilter(stateValue)}>{displayName}</div>
+      <Link to="/campaigns" className={className} onClick={() => {this.setCampaignStateFilter(stateValue)}}>{displayName}</Link>
```
Calanthe: Isn't `onClick={this.setCampaignStateFilter(stateValue)}` simpler?

public/styles/components/sidebar/_sidebar.scss
```diff
@@ -105,6 +105,11 @@
 
   padding: 5px 40px;
 
+  cursor: pointer;
```
Calanthe, line 108: I've done the same in my PR ;) https://github.com/guardian/campaign-central/pull/45
Will get rid of it.

public/components/Sidebar/Sidebar.js
lps88: It would have to be `onClick={this.setCampaignStateFilter.bind(this, stateValue)}` and I prefer it how it is now! I didn't mean to add those extra brackets though

Calanthe: ok, let's leave it as it is now

Calanthe:
👍

PR #49, opened 2016-11-10
Sw cta ctr

steppenwells:
Show the number of clicks and the CTR for call to actions:

screen shot 2016-11-10 at 15 48 21

If there are multiple CTAs a totals row is also displayed.

### DIFF
--- a/app/model/CallToAction.scala
+++ b/app/model/CallToAction.scala
@@ -4,7 +4,8 @@ import ai.x.play.json.Jsonx
 import play.api.libs.json.Format
 
 case class CallToAction (
-                        builderId: Option[String] // the id from the call to action builder
+                        builderId: Option[String], // the id from the call to action builder
+                        trackingCode: Option[String] = None
                         )
 
 object CallToAction {
--- a/app/repositories/GoogleAnalytics.scala
+++ b/app/repositories/GoogleAnalytics.scala
@@ -2,46 +2,22 @@ package repositories
 
 import java.util.concurrent.Executors
 
-import ai.x.play.json.Jsonx
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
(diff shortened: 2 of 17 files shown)


app/model/CallToAction.scala
```diff
@@ -4,9 +4,10 @@ import ai.x.play.json.Jsonx
 import play.api.libs.json.Format
 
 case class CallToAction (
-                        builderId: Option[String] // the id from the call to action builder
+                        builderId: Option[String], // the id from the call to action builder
+                        trackingCode: Option[String] = None
```
kelvin-chappell, line 8: Does `trackingCode` need a default value?

app/repositories/GoogleAnalytics.scala
```diff
@@ -315,6 +286,94 @@ object GoogleAnalytics {
...
+
+        Future{
+          Logger.debug(s"async refresh of CTA clicks for campaign $campaignId")
+          fetchAndStoreCtaClicksForCampaign(campaignId)
+        } // serve stale but spawn refresh future
+        Some(report)
+      }
+      case Miss => {
+        Logger.debug(s"getting CTA clicks for campaign $campaignId - cache miss fetching sync")
+        fetchAndStoreCtaClicksForCampaign(campaignId)
```
kelvin-chappell, line 310: If you miss, this will block until the data is fetched.  Is the aim to run `getCtaClicksForCampaign` as a background job?

kelvin-chappell (approved):
👍  Just a couple of Qs

app/repositories/GoogleAnalytics.scala
steppenwells: Yep, in normal operation the first hit is syncronous and from there serving stale kicks in. There will be a job that will find stale cache entries and initiate the refresh. There is a management API for this but the scheduled job still needs to be implemented.

PR #51, opened 2016-11-11
show how a campaign is doing at a glance on the list view

steppenwells:
Used the analytics summary to display how a campaign is doing against where it should be and provides traffic light style colouring.

screen shot 2016-11-11 at 16 29 14

Hopefully it's clear. i plan to roll this out and then gather feedback on wording etc once it's in front of people.


kelvin-chappell (approved):
👍

PR #53, opened 2016-11-16
Show spinner while waiting for traffic driver data to load

kelvin-chappell:
So that you know you're waiting for a result, rather than there being no data available.

This also removes some repeated code in the traffic driver stats template.


Calanthe:
👍

PR #54, opened 2016-11-16
Look up DFP line items in multiple orders

kelvin-chappell:
There are separate orders for hosted and paid traffic drivers.  That's why the expected traffic drivers weren't appearing in paid campaigns.


steppenwells (approved):
Looks sensible

👍

PR #56, opened 2016-11-28
Campaign list - column's sorting

Calanthe:
This PR introduces column's sorting on the campaign list.

screen shot 2016-11-28 at 17 33 57

@clloyd @steppenwells

### DIFF
--- a/public/components/CampaignList/CampaignList.js
+++ b/public/components/CampaignList/CampaignList.js
@@ -11,33 +11,39 @@ class CampaignList extends React.Component {
     campaigns: []
   };
 
-  sortOrder = {
-    'name': true, //false = DESC, true = ASC
-    'type': false,
-    'status': false,
-    'actualValue': false,
-    'startDate': false,
-    'endDate': false
-  };
+  getSortOrder = (column) => {
+    let order = false; //false = DESC, true = ASC
+    let DOMNode = this.refs['sort-' + column];
+    let cssClass = DOMNode ? DOMNode.className : '';
 
-  setHeaderCssClass = (currentHeader) => {
-    let sortedColumn = this.props.campaignSortColumn || 'name';
-    let order = this.sortOrder[sortedColumn] ? 'asc' : 'desc';
-    let cssClass = 'campaign-list__header-order';
-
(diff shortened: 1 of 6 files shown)


public/components/CampaignList/CampaignList.js
```diff
@@ -11,6 +11,40 @@ class CampaignList extends React.Component {
...
+    sortOrder: {
+      'name': true, //false = DESC, true = ASC
+      'type': false,
+      'status': false,
+      'actualValue': false,
+      'startDate': false,
+      'endDate': false
+    },
+    headersCssClasses: {
+      'name': 'campaign-list__header-order--desc name',
```
Calanthe: I know that you prefer to have one css class per tag @clloyd. Do you think that's an issue here?

clloyd: Does any of this need to be stored in the component state? Can it be implied from the redux global state? Same from the classes, can they be constructed on each render, then you don't need to keep them all in sync?

public/components/CampaignList/CampaignList.js
```diff
@@ -11,6 +11,40 @@ class CampaignList extends React.Component {
...
+  };
+
+  setCampaignSort = (c) => {
+    const order = this.state.sortOrder[c] ? 'asc' : 'desc';
+
+    //sort campaign list
+    this.props.uiActions.setCampaignSort(c, this.state.sortOrder[c]);
+    //reset all of the headers' css classes so by default there is always up and down arrow
+    for (let value in this.state.headersCssClasses) {
+      this.state.headersCssClasses[value] = 'campaign-list__header-order ' + value;
```
clloyd: I don't think you need any of this information in the component state... but even if you do, don't modify this.state directly it won't trigger a render, you should call this.setState

public/components/CampaignList/CampaignList.js
clloyd: I've no huge issue with two class names where it makes sense, I think it does in this case.

clloyd (changes requested):

clloyd (approved):

PR #57, opened 2016-11-29
Possibly improve measurement of campaign 'uniques'

kelvin-chappell:
This change measures unique page views per campaign directly from GA, instead of deriving the figure from unique page views per campaign page.
This value shouldn't be sampled, so should be accurate.
But it might not be what people expect!

Although assuming happiness ≈ accuracy / expectation * randomFactor.

There's some refactoring too.

/cc @steppenwells @guardian/labs-beta

### DIFF
--- a/app/controllers/CampaignApi.scala
+++ b/app/controllers/CampaignApi.scala
@@ -1,7 +1,5 @@
 package controllers
 
-import java.util.UUID
-
 import model._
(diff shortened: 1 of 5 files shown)


kelvin-chappell:
Uniques for the campaign is the same as uniques per page summed, so redundant!

PR #58, opened 2016-12-02
Sw refresh bug fixes

steppenwells:
Fed up with not making progress on unique users I've fix a few bugs that I found in trello. They are:

- after refresh ended campaigns become live again
- after refresh ended campaigns loose their logo
- rename capi import to import campaign

And as a bonus stop saying ended campaigns end in -x days, not say they ended x days ago (and depluralise "days" when there's only one day in question).


kelvin-chappell (approved):
👍

PR #59, opened 2016-12-05
Campaigns' columns sorting fixes

Calanthe:
The campaign's list by default should be sorted by the end date field. All 'not yet configured values' should be infinite so by default (DESC order) they will be at the bottom of the list.

I also fixed the `min-width` value of columns so they will be correctly visible on smaller breakpoints.

Before:
screen shot 2016-12-05 at 15 01 43

After:
screen shot 2016-12-05 at 15 01 13


steppenwells (approved):
Looks god 👍

PR #60, opened 2016-12-05
Kill webpack dev server on sbt shutdown

kelvin-chappell:
Have to keep manually killing the webpack server whenever I restart sbt.
This change does it as part of the script.


steppenwells (approved):
:+1:

PR #61, opened 2016-12-07
Fetch traffic driver suggestions for a campaign

kelvin-chappell:
This PR just provides an api endpoint to fetch the suggestions.
It runs through a list of increasingly relaxed fetches until it arrives at one that has at least one result, so that there are more likely to be false positives than missing values.

I haven't filtered out the traffic drivers that are already linked to a campaign yet because this will be valuable to see if the fetching queries are sensible.

The next steps will be to integrate this into the campaign page, and then link the accepted line items to the campaign in DFP.

/cc @guardian/labs-beta


steppenwells (approved):
👍

PR #62, opened 2016-12-09
Link DFP line items to campaigns

kelvin-chappell:
This provides an API endpoint to link a traffic driver to a campaign.

Next step is to use this endpoint in the campaign page, and integrate it with the suggestions functionality in #61.

/cc @guardian/labs-beta

PR #65, opened 2016-12-13
Show traffic driver suggestions

kelvin-chappell:
Eg.
image

Buttons have no effect until linking code in #62 is deployed.

/cc @guardian/labs-beta

### DIFF
--- a/app/model/TrafficDriver.scala
+++ b/app/model/TrafficDriver.scala
@@ -57,7 +57,7 @@ object TrafficDriver {
     TrafficDriver(
       id = lineItem.getId,
       name = lineItem.getName,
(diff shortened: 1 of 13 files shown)


steppenwells (changes requested):
Would it be possible to wire up the buttons to real functionality before we roll this out?

Also how is ignore suggestion going to work?

kelvin-chappell:
Yes, I'll keep this on hold until the linking is actually working.

I haven't worked out what to do with the ignore yet! Store it in dynamo I should think.

kelvin-chappell:
The 'accept suggestion' functionality is now working!

(I'll get the 'ignore suggestion' working in a separate PR.)

kelvin-chappell:
@steppenwells @clloyd Do you mind having a look over commit 0a23aef0a82d0f1506de9f72d3014d670d6bea8f please?

steppenwells (approved):
looks legit

PR #67, opened 2016-12-14
Sw hmac on management

steppenwells:
Adds HMAC authentication on management api - see https://github.com/guardian/panda-hmac for 'details'

Adds a new endpoint to refresh expiring campaigns form capi, this will either update them to the dead state or prolong them if they have been extended in tag manager.


kelvin-chappell (approved):
Looks good to me

PR #71, opened 2016-12-22
Bump DFP client version

kelvin-chappell:
And:
* explicitly include guava version, because transitive dependency gets evicted
* bump scalatest
* remove redundant joda-convert dependency


steppenwells (approved):
:+1:

PR #74, opened 2017-01-03
Filter campaigns with query parameters

lps88:
From [Campaign Central Feedback](https://docs.google.com/document/d/1nJZEdfjYuouF9yjwmmJnlNUTD70VnOtIc9xK-QKg0sY/edit) document: 

__Bookmarkable or linkable URLs__
"Please always update URL when clicking on the menu so it can be bookmarked and navigation (forward backward) is not broken"

__To do__: remove old filter mechanism if this is accepted as a better alternative.

CC: @steppenwells @kelvin-chappell


kelvin-chappell (approved):
Looks good to me

PR #75, opened 2017-01-04
Remove old filtering

lps88:
- Remove redux actions no longer needed after implementing filtering with query parameters, see https://github.com/guardian/campaign-central/pull/74

- Implemented change to default campaigns view as per feedback document:

__Change homepage to just include live campaigns__
"Because this is the campaigns we care about the most and most analysed" - Rob

- Tweak campaign list logos sizing/positioning to be more compact:

"Remove or reduce logos sizes so we can get more in" - Rob


kelvin-chappell (approved):
👍

PR #76, opened 2017-01-04
Campaigns list updates

lps88:
- Separate 'target' into its own column
- Separate 'days left' into its own column
- Format dates properly

Before:

image

After:

image


kelvin-chappell (approved):
👍

PR #77, opened 2017-01-04
Campaigns sorting

lps88:
Fixes the default sorting (sort by end date on load wasn't working) and simplifies sort functions a bit.

image

Also stop the error bar from spilling off the screen:

image

CC: @steppenwells @kelvin-chappell


kelvin-chappell (approved):
👍

PR #79, opened 2017-01-05
Campaigns list - simplify uniques column

lps88:
Instead of showing all the info, just show uniques, put the rest into a tooltip sort of thing.

image


steppenwells (approved):
Looks really nice

PR #80, opened 2017-01-05
return an empty targets report if target and run rate can't be calcul…

steppenwells:
…ated

This was 404 ing which gave an ugly error message and it prevented the analytics charts displaying (which is worse)


lps88 (approved):
🎉

PR #82, opened 2017-01-05
CampaignList sorting overhaul

lps88:
- Use `props.campaignSortOrder` to determine current sort order rather than css classes
- Use `1`/`-1` as values of `campaignSortOrder` instead of `true`/`false`
- Reduce duplication in column headers mark-up
- Fix styling of arrows so they don't need to each be positioned separately
- Get rid of unnecessary mixins etc in CSS

CC @guardian/labs-beta


steppenwells (approved):
👍

PR #84, opened 2017-01-10
Enable rejection of traffic driver suggestions

kelvin-chappell:
This change keeps a persistent store of rejected traffic driver suggestions so that they don't keep coming up.

/cc @guardian/commercial-dev

### DIFF
--- a/cloudformation/dynamoDb.json
+++ b/cloudformation/dynamoDb.json
@@ -197,6 +197,46 @@
         }
       },
       "DeletionPolicy": "Retain"
+    },
+    "TrafficDriverRejectDBTable": {
+      "Type": "AWS::DynamoDB::Table",
+      "Properties": {
+        "AttributeDefinitions": [
+          {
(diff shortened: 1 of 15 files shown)


cloudformation/dynamoDb.json
```diff
@@ -197,6 +197,46 @@
...
+            "AttributeName": "campaignId",
+            "KeyType": "HASH"
+          },{
+            "AttributeName": "lineItemId",
+            "KeyType": "RANGE"
+          }
+        ],
+        "ProvisionedThroughput": {
+          "ReadCapacityUnits": "1",
+          "WriteCapacityUnits": "1"
```
steppenwells: I know times are tight but we can probably splurge on a bit more capacity. Setting it to 1 is probably a recipe for alarms firing and throttling on standard use. I'd suggest 5 at the minimum

steppenwells (approved):
Looks good

Small comment on the dynamo provisioning.

PR #85, opened 2017-01-10
Sw qualified report

steppenwells:
adds a qualified percentage report.

screen shot 2017-01-10 at 16 50 54

show video completion stats and dwell times > 15s for other content types. Only shows stats for types that are available.

This data suffers form sampling, we should keep an eye on it to check it's still useful.

### DIFF
--- a/app/model/reports/QualifiedPercentagesReport.scala
+++ b/app/model/reports/QualifiedPercentagesReport.scala
@@ -51,10 +51,10 @@ object QualifiedPercentagesReport {
     }
   }
 
-  val qualifiedMetrics = Map(
-    "articleDwellTime" -> ContentTypeDwellTimeMetric("article", 15),
-    "galleryDwellTime" -> ContentTypeDwellTimeMetric("gallery", 15),
-    "interactiveDwellTime" -> ContentTypeDwellTimeMetric("interactive", 15)
+  val qualifiedMetricFetchers = List(
+    ContentTypeDwellTimeMetric("article", 15),
+    ContentTypeDwellTimeMetric("gallery", 15),
+    ContentTypeDwellTimeMetric("interactive", 15)
   )
 
   def generateReport(campaignId: String): Option[QualifiedPercentagesReport] = {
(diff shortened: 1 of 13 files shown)


app/model/reports/QualifiedPercentagesReport.scala
```diff
@@ -0,0 +1,131 @@
...
+
+case class QualifiedPercentagesReport(campaignId: String, metrics: Map[String, QualifiedMetricReport]) {
+
+  def refresh = QualifiedPercentagesReport.generateReport(campaignId)
+
+}
+
+object QualifiedPercentagesReport {
+
+  implicit val ec = AnalyticsDataCache.analyticsExectuionContext
```
kelvin-chappell, line 26: Typo here

app/model/reports/QualifiedPercentagesReport.scala
```diff
@@ -0,0 +1,131 @@
...
+          report.refresh
+        } // serve stale but spawn refresh future
+        Some(report)
+      }
+      case Miss => {
+        Logger.debug(s"getting qualified percentages for campaign $campaignId - cache miss fetching sync")
+
+        generateReport(campaignId)
+      }
+    }
```
kelvin-chappell, line 51: I wonder if we can reuse this caching pattern match as it seems to recur with slight variations?  But the variations may make it not worthwhile.

app/repositories/GoogleAnalytics.scala
```diff
@@ -140,6 +140,103 @@ object GoogleAnalytics {
...
   }
 
+  // qualified reporting functions
+
+  def loadTotalCampaignContentTypeViews(campaignFilter: String, contentType: String, startDate: DateTime, endDate: Option[DateTime]): Long = {
+    loadTotalViewsWithFilter(s"$campaignFilter;ga:dimension5==$contentType", startDate, endDate)
+  }
+
+  def loadCampaignContentTypeViewsWithDwellTime(campaignFilter: String, contentType: String, qualifiedDwellTime: Int, startDate: DateTime, endDate: Option[DateTime]): Long = {
+    loadTotalViewsWithFilter(s"$campaignFilter;ga:dimension5==$contentType;ga:timeOnPage>=$qualifiedDwellTime", startDate, endDate)
```
kelvin-chappell, line 150: This is more straightforward than I was expecting.

public/components/CampaignAnalytics/Analytics/CampaignQualifiedChart.js
```diff
@@ -1,17 +1,85 @@
...
+    var videoComplete = this.props.qualifiedReport.metrics.videoComplete;
+    if (videoPlays) {
+      data.push( {name: "video complete", percentage: videoComplete.percentage, fill: getFillColour(index) } );
+    }
+
+    return data;
+  }
+
+  formatPercent(p) {
+    return parseFloat(p).toFixed(2) + '%';
```
kelvin-chappell, line 55: I'm not sure if this is also the format for traffic driver CTRs?

kelvin-chappell (approved):
👍 
Not sure what to do about sampling but these are likely to be low numbers, especially for video completions.  Hopefully indicative at least.

PR #86, opened 2017-01-12
Sw logo ctr

steppenwells:
Shows the number of clicks on the sponsor logo and click through rate.

NB the data for this has only recently started being recorded, and the data is sampled. This is likely to lead to some very underwhelming counts and ctrs (but at least they'll be round).

screen shot 2017-01-12 at 15 57 56


kelvin-chappell (approved):
👍

PR #88, opened 2017-01-16
Make traffic driver end-date optional

kelvin-chappell:
Currently traffic drivers are assumed to have an end-date, as it's assumed that they will end when the campaign ends, but this may cause no traffic drivers to appear.  So this is a fix.


steppenwells (approved):
👍

PR #89, opened 2017-01-16
Sw import finished campaign

steppenwells:
Uses data from the sponsorship object not the CAPI data to derive sponsor name. This means that expired campaigns should be imported correctly.


kelvin-chappell (approved):
👍

PR #91, opened 2017-07-12
update DFP dependency version

gidsg:
Update DFP dependency to latest version; as v201608 is being deprecated as per http://googleadsdeveloper.blogspot.co.uk/2017/07/sunset-of-dfp-api-v201608.html
CC @guardian/commercial-dev


kelvin-chappell (approved):
👍

PR #92, opened 2017-07-13
Fix JS SyntaxError so that webpack stops failing

katebee:
Webpack was unhappy:

picture 836

This PR removes the `null`s that were making Webpack throw syntax errors.

Campaign Central will run again! 🎉


gidsg (approved):

kelvin-chappell (approved):
Thanks Kate 👍

PR #93, opened 2017-07-13
Ensure Webpack throws a non-zero exit code... Bonus: add emoji 😎 

katebee:
## Webpack Fix

- add the `--bail` flag to the Webpack command

We recently had an issue that caused `sh scripts/setup.sh` to explode, however the tests would still pass on CircleCI.

This appears to be the most elegant way to ensure that Webpack will throw a non-zero exit code, and cause CircleCI to fail if there are errors encountered. More information here: https://github.com/webpack/webpack/issues/708

If I use this flag, add back the issue from #92, and start opening a PR, I will now see the tests failing on CircleCI:

picture 838

Also, you should now see `Exit status 1` on the CLI accordingly if something breaks:

picture 837

I am not adding the flag to build-dev, since we probably don't want to bail out of watch mode if a developer is hacking around on their local version.

## Spelling and Emoji

- Fixing slight spelling and casing issue in setup.sh 

- I like to meddle... so bonus emoji! Everything is better with emoji ❤️✌️

Running `sh scripts/setup.sh` will now print something like this to the terminal:

picture 839

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "description": "Tool for managing campaigns",
   "scripts": {
-    "build": "webpack -p ./public/app.js ./public/build/app.js --config ./public/build_config/webpack.prod.conf.js",
+    "build": "webpack -p ./public/app.js ./public/build/app.js --config ./public/build_config/webpack.prod.conf.js --bail",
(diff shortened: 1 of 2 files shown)


kelvin-chappell (approved):
👍

PR #94, opened 2017-07-19
Refactor error-handling

kelvin-chappell:
So that errors are not thrown but handled as part of the method response.


LATaylor-guardian:
👍

PR #96, opened 2017-07-24
add note to readme about nginx error

gidsg:
Adding a note to readme as @katebee and I both hit this error.


LATaylor-guardian:
👍

PR #98, opened 2017-07-25
Remove campaign analytics graphs.

LATaylor-guardian:
I added logging to check that nobody was using these which confirmed our assumptions. They also are expensive to load and don't add any known value.


kelvin-chappell (approved):
Looks good.  The job that generates the data behind the graphs will need to be removed as well, and perhaps the tables.

public/components/CampaignAnalytics/CampaignAnalytics.js
```diff
@@ -1,97 +0,0 @@
...
-
-    if (!this.props.campaignPageViews) {
-      return <div className="campaign-info__body">Loading... </div>;
-    }
-
-    return (
-      <div>
-        <div className="campaign-info__body">
-
-          <CampaignPerformanceSummary campaign={this.props.campaign}
```
janua, line 47: Do we not want to keep CampaignPerformance? Just change the graph? Or maybe your plan is to completely re-write the graph?

LATaylor-guardian:
> Looks good. The job that generates the data behind the graphs will need to be removed as well, and perhaps the tables.

Yes, I'll do that at a later date when we no for sure no one cares about these.

public/components/CampaignAnalytics/CampaignAnalytics.js
LATaylor-guardian: Just planned to recreate (using other graphs as a reference)

PR #101, opened 2017-07-26
Bump Play to version 2.6

kelvin-chappell:


kelvin-chappell:
[Upgrade Campaign Central to Play 2.6](https://trello.com/c/XTJ7Ecss/123-upgrade-campaign-central-to-play-26)

LATaylor-guardian:
@kelvin-chappell looks okay to me but I'll default to @philwills given he done this for Riff Raff recently.

app/di.scala
```diff
@@ -2,19 +2,27 @@ import controllers._
...
-class AppComponents(context: Context) extends BuiltInComponentsFromContext(context) with AhcWSComponents {
+class AppComponents(context: Context)
+  extends BuiltInComponentsFromContext(context)
+    with AhcWSComponents
+    with AssetsComponents
+    with HttpFiltersComponents {
+
+  // disabling some default filters until value is clear in this context
+  override def httpFilters: Seq[EssentialFilter] =
+    super.httpFilters.filterNot(filter => filter == csrfFilter || filter == allowedHostsFilter)
```
philwills, line 25: If I'm reading this right, this disables the CSRFFilter. That seems like a bad idea to me unless there's a really compelling reason.

philwills (changes requested):
With the exception of the filters, this looks reasonable.

app/di.scala
kelvin-chappell: My thinking is that the app will be no less secure than it is now with these filters disabled, it's clear that they are disabled, and I wanted to make sure that I haven't broken anything before enabling these filters one by one.  The CSRF filter looks like there will be quite a few changes involved.

philwills (approved):

app/di.scala
philwills: That makes sense, I had thought it was on by default in 2.5.

PR #102, opened 2017-07-26
Update cloudformation for OphanExcavatorDynamoPolicy

janua:
This adds `dynamodb:Scan` and also adds the `campaigns` table from campaign central as we need to scan this for the campaigns.


LATaylor-guardian:
👍

PR #106, opened 2017-07-28
Provide API endpoints for uniques and pageviews per campaign.

shtukas:


app/model/CampaignUniquesItem.scala
```diff
@@ -0,0 +1,32 @@
...
+                                   uniques: Long
+                                 ){
+  def toItem = Item.fromJSON(Json.toJson(this).toString())
+}
+
+object CampaignUniquesItem {
+  implicit val CampaignUniquesItemFormat: Format[CampaignUniquesItem] = Jsonx.formatCaseClass[CampaignUniquesItem]
+  def fromJson(json: JsValue) = json.as[CampaignUniquesItem]
+
+  def fromItem(item: Item) = try {
```
janua, line 24: Maybe we should think about switching to using Scanamo?

janua:
LGTM 👍

app/model/CampaignUniquesItem.scala
LATaylor-guardian: Yeah I have it written down. We should also use Circe for json - much cleaner and less boilerplate.

PR #107, opened 2017-07-28
Remove Dependency on GA and any unecessary code.

LATaylor-guardian:
We plan on using the data lake to source our data rather than from GA, which Campaign Central is currently using. This allows us to drastically reduce the amount of code and complexity within this app for the following reasons:

1. No one is using Campaign Central. No really... we've heard this from users that the data is wrong and we've put logging in ourselves to confirm this - no one is signing in.
2. A substantial amount of code was written in order to support a substantial number of graphs within the UI. These were not valuable to users.
3. Given the use of the GA api, the application had to make queries against the data source, perform some logic, aggregation and type coercion and then write the results of that to Dynamo as it's caching layer. By switching to retrieving our data from the data lake we can avoid this. We run a job on the excavator, which can be run periodically and scale as needed in order to perform the aggregations we require and then write off those pre-aggregated results to the data store that Campaign Central uses. Pre-aggregating, especially across very large datasets is a common strategy and serves us well for our needs here. You can see from the campaign central code that it is having to work around the limitation of not being able to aggregate across the entire dataset of a campaign. Any heavy lifting we want done by the Excavator on the EMR cluster where it's built for scale while keeping Campaign Central as a dumb app which makes DB calls for that data and presents it to the user.

There are further improvements here such as how we import/create campaigns. I'm also sure there is more code that may be deleted that I have missed.

By simplifying the codebase and removing the unnecessary we can now begin to focus on presenting our new data in a way that our stakeholders find valuable and of use. We will work closely with them to do this.


shtukas:
:+1:

JonNorman (approved):
Looks good. To me.

janua:
Great! :+1:

LATaylor-guardian:
This has a few conflicts now so will recreate.

PR #105, opened 2017-07-27
Bump Play to version 2.6 again

kelvin-chappell:
This is a another attempt at #101 
But this time using an up-to-date logback configuration.

PR #108, opened 2017-08-02
Revert "Bump Play to version 2.6 again"

LATaylor-guardian:
Reverts guardian/campaign-central#105

@kelvin-chappell I'm reverting this as it's causing some client side requests from campaign central to fail.


kelvin-chappell:
Thanks @LATaylor-guardian I'll get this sorted out!

LATaylor-guardian:
@kelvin-chappell I had a quick look and noticed that if I didn't specify a content-type on the client side requests it worked. So it looks to be a problem accepting the application/json content type.

kelvin-chappell:
Yes when the ajax request is made it has text/html content type but when you make the request direct it's application/json.  So it should be quite simple to fix.
With play 2.5 it was application/json in both cases.

PR #109, opened 2017-08-02
Add graph showing uniques for campaign.

LATaylor-guardian:
picture 118


janua:
LGTM 👍

LATaylor-guardian:
This isn't perfect but I'm going to merge and we can improve as we notice problems with it when using real data.

PR #110, opened 2017-08-03
Upgrade to Play 2.6 attempt 3

kelvin-chappell:
This is the same as the last attempt apart from the addition of 1c3e5c1
This time the ajax get requests don't have content-types because they make no sense for get requests, and Play madly tries to parse the body.


LATaylor-guardian:
👍

PR #112, opened 2017-08-03
Fix div by zero error

kelvin-chappell:
Seeing here: https://campaign-central.gutools.co.uk/api/v2/campaigns/d6da8e74-865b-4d8e-9d7f-654df21211cc/uniques


LATaylor-guardian:
Thanks! 👍

PR #113, opened 2017-08-03
Fix JS error mapping null

kelvin-chappell:
I was too hasty with the last fix: #112 
It just pushed the error on a bit.

### DIFF
--- a/public/components/CampaignAnalytics/Analytics/CampaignUniquesChart.js
+++ b/public/components/CampaignAnalytics/Analytics/CampaignUniquesChart.js
@@ -9,16 +9,17 @@ export default class CampaignUniquesChart extends React.Component {
 
   render () {
 
-    console.log(this.props.data);
-
-    let data = this.props.data.map( (item) => {
+    let data = [];
+    if (this.props.data) {
+      data = this.props.data.map((item) => {
         return {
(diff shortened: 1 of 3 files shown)


janua:
:+1:

public/components/CampaignAnalytics/Analytics/CampaignUniquesChart.js
```diff
@@ -9,16 +9,17 @@ export default class CampaignUniquesChart extends React.Component {
 
   render () {
 
-    console.log(this.props.data);
-
-    let data = this.props.data.map( (item) => {
+    let data = [];
+    if (this.props.data) {
```
LATaylor-guardian, line 13: I'm not sure we want to show the graph if there is no data. Can we add:
```
if (!this.props.data) return null;
```
as the first line of the render method instead?

kelvin-chappell: @LATaylor-guardian Just saw your comment a second too late!  I'll do it in another PR

LATaylor-guardian: @kelvin-chappell of course, no worries :)

PR #115, opened 2017-08-04
Update case classes

janua:
Remove `startDate` and `endDate` from `CampaignPageViewsItems` and `CampaignUniquesItem`.

See https://github.com/guardian/ophan-data-lake/pull/1064.


LATaylor-guardian:
👍

PR #116, opened 2017-08-07
Fix display of dates for campaigns screen.

LATaylor-guardian:


public/components/CampaignList/CampaignListItem.js
```diff
@@ -27,8 +27,8 @@ class CampaignListItem extends React.Component {
     const totalUniques = this.props.analyticsSummary.totalUniques;
     const targetToDate = this.props.analyticsSummary.targetToDate;
     const now = new Date();
-    const endDate = this.props.campaign.endDate;
-    const startDate = this.props.campaign.startDate;
+    const endDate = new Date(this.props.campaign.endDate);
```
kelvin-chappell, line 30: What was the type before?

kelvin-chappell (approved):
👍

public/components/CampaignList/CampaignListItem.js
LATaylor-guardian: A String by the looks of it. Not sure how it ever worked.

PR #117, opened 2017-08-07
Small UI improvements.

LATaylor-guardian:
- Remove unnecessary UI info.
 - Don't show content performance graph if data is an empty array.


kelvin-chappell (commented):
👍

PR #120, opened 2017-08-09
Retrieve yesterdays data for latest uniques.

LATaylor-guardian:
Now that the date format is consistently `YYYY-MM-dd` we can look for items that equal yesterdays date in this format.


kelvin-chappell (approved):
Looks right to me 👍

PR #121, opened 2017-08-11
Add table to store latest analytics data per campaign.

LATaylor-guardian:
This table will store the latest number of uniques and pageviews for a campaign. The data lake job will overwrite this data per campaign when it runs every night. This greatly simplifies our ability to retrieve this information from Campaign Central and allows us to make use of an existing job that is already running.


shtukas (approved):

LATaylor-guardian:
Created tables. Merging.

PR #126, opened 2017-08-18
New route for referrals from on-platform

kelvin-chappell:
At the moment the CampaignReferral model assumes that all referrals come from cards in containers, but I foresee this including freestanding cards served with content or from the ad-server, and also email referrals.  In principle it includes app traffic as well as web traffic but need to tweak query to get app data in and then there will be more options for platform.

Going to do rendering in a separate PR.

### DIFF
--- a/app/model/CampaignReferral.scala
+++ b/app/model/CampaignReferral.scala
@@ -1,0 +1,79 @@
+package model
+
+import java.time.LocalDate
+
+import com.amazonaws.services.dynamodbv2.document.Item
+import play.api.Logger
+import play.api.libs.json._
+import repositories.CampaignReferralRepository
+
+import scala.util.control.NonFatal
+
+// A referral to any item in a campaign from an on-platform source in a particular period
+case class CampaignReferral(
+  campaignId: String,
+  platform: String,
+  edition: String,
+  path: String,
+  numClicks: Int,
+  containerIndex: Int,
+  containerName: String,
+  cardIndex: Int,
+  cardName: String,
+  firstReferral: LocalDate,
+  lastReferral: LocalDate
+)
+
+object CampaignReferral {
+
+  // dynamodb item -> CampaignReferral
+  implicit val itemReads: Reads[CampaignReferral] = Reads { json =>
+    val platformUrlComponent: Seq[String] = (json \ "platformUrlComponent").as[String].split("\\|").toSeq
+
+    def platform(s: String) = s match {
+      case "NEXT_GEN" => "Web"
+      case _          => "unknown"
+    }
+
+    def ifPresent(s: Option[String]): String = s.map(_.trim) getOrElse "unknown"
+
+    def indexIfPresent(s: Option[String], prefix: String): Int = s.map(_.trim.stripPrefix(prefix).toInt) getOrElse -1
+
+    def asDate(js: JsLookupResult): LocalDate = LocalDate.parse(js.as[String])
+
+    JsSuccess(
+      CampaignReferral(
+        campaignId = (json \ "campaignId").as[String],
+        platform = ifPresent(platformUrlComponent.headOption.map(platform)),
+        edition = ifPresent(platformUrlComponent.lift(3)),
+        path = ifPresent(platformUrlComponent.lift(4)),
+        numClicks = (json \ "clicks").as[Int],
+        containerIndex = indexIfPresent(platformUrlComponent.lift(5), "container-"),
+        containerName = ifPresent(platformUrlComponent.lift(6)),
+        cardIndex = indexIfPresent(platformUrlComponent.lift(8), "card-"),
+        cardName = ifPresent(platformUrlComponent.lift(9)),
+        firstReferral = asDate(json \ "firstReferral"),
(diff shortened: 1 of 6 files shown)


kelvin-chappell:

conf/routes
```diff
@@ -40,6 +40,7 @@ PUT /api/campaigns/:id/driver/:driverId     controllers.CampaignApi.acceptSugges
 PUT /api/campaigns/:id/not-driver/:driverId controllers.CampaignApi.rejectSuggestedCampaignTrafficDriver(id: String, driverId: Long)
 GET /api/campaigns/:id/driverstats          controllers.CampaignApi.getCampaignTrafficDriverStats(id: String)
 GET /api/campaigns/:id/ctastats             controllers.CampaignApi.getCampaignCtaStats(id: String)
+GET /api/campaigns/:id/referrals            controllers.CampaignApi.getCampaignReferrals(id: String)
```
LATaylor-guardian: I've done `/api/v2` for now just so we mark the new endpoints. Then when we eventually clear all the old ones out we can remove the version extension.

app/model/CampaignReferral.scala
```diff
@@ -0,0 +1,79 @@
...
+
+object CampaignReferral {
+
+  // dynamodb item -> CampaignReferral
+  implicit val itemReads: Reads[CampaignReferral] = Reads { json =>
+    val platformUrlComponent: Seq[String] = (json \ "platformUrlComponent").as[String].split("\\|").toSeq
+
+    def platform(s: String) = s match {
+      case "NEXT_GEN" => "Web"
+      case _          => "unknown"
```
janua, line 35: Why not cover all the cases?

LATaylor-guardian:
👍

app/model/CampaignReferral.scala
kelvin-chappell: That's all I'm getting at the moment.  I need to tweak my query once I've got the whole thing working end-to-end.

conf/routes
kelvin-chappell: OK.  I'll v2 it.

app/model/CampaignReferral.scala
```diff
@@ -0,0 +1,79 @@
...
+      CampaignReferral(
+        campaignId = (json \ "campaignId").as[String],
+        platform = ifPresent(platformUrlComponent.headOption.map(platform)),
+        edition = ifPresent(platformUrlComponent.lift(3)),
+        path = ifPresent(platformUrlComponent.lift(4)),
+        numClicks = (json \ "clicks").as[Int],
+        containerIndex = indexIfPresent(platformUrlComponent.lift(5), "container-"),
+        containerName = ifPresent(platformUrlComponent.lift(6)),
+        cardIndex = indexIfPresent(platformUrlComponent.lift(8), "card-"),
+        cardName = ifPresent(platformUrlComponent.lift(9)),
```
janua, line 54: What sort of string exists in here? Will `cardName` always be in 9th position for example? Don't know the string!

app/model/CampaignReferral.scala
```diff
@@ -0,0 +1,79 @@
...
+    val (unknown, known) = CampaignReferralRepository.getCampaignReferrals(campaignId) partition (_.path == "unknown")
+    unknown foreach { referral =>
+      Logger.warn(s"Unknown referral: $referral")
+    }
+    known
+  }
+
+  def fromItem(item: Item): Option[CampaignReferral] =
+    try {
+      Some(Json.parse(item.toJSON).as[CampaignReferral])
```
janua, line 73: Why aren't we using Scanamo? A lot of the boilerplate in this PR will disappear from this PR if we used it directly with the case class.

LATaylor-guardian: We need to use Scanamo and Circe - we just haven't got round to it yet. Probably worth doing once we've got the core data in the app and we begin to do some cleanup.

app/model/CampaignReferral.scala
kelvin-chappell: At the moment the 9th field is like `card-3` but I expect this parsing logic to change a lot when it includes freestanding cards and ad-served sources etc

app/model/CampaignReferral.scala
kelvin-chappell: No reason not to use Scanamo except that no-one working on this project before had ever used it.
I don't know what Circe is.

app/model/CampaignReferral.scala
kelvin-chappell: Once I have a more stable data structure can do the parsing in the excavator instead and just read a case class from dynamo.

app/model/CampaignReferral.scala
LATaylor-guardian: @kelvin-chappell You'll like Circe: https://github.com/circe/circe - its a newish but very popular json library which  does fully automatic derivation. Will reduce the boilerplate in this project allot.

janua: @kelvin-chappell Scanamo is great; it will turn your Dynamo query into one line (ish)

PR #128, opened 2017-08-21
Add on-platform referrals section to campaign page

kelvin-chappell:
This is the first attempt at a view of referrals:
image

Needs quite a lot of work after this PR:
* Show the full dataset for the campaign (currently just a few days, and first and last referral dates are misleading)
* Show impression count and CTR as well as click count
* Aggregate data so not quite so fine-grained (probably don't need to separate out editions, but will find out what is useful)
* Add more platforms
* Add ad-served cards as well for comparison
* Add emails

### DIFF
--- a/app/model/CampaignReferral.scala
+++ b/app/model/CampaignReferral.scala
@@ -1,5 +1,6 @@
 package model
 
+import java.math.BigInteger
 import java.time.LocalDate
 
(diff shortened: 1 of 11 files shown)


kelvin-chappell:

janua:
LGTM 👍

PR #129, opened 2017-08-23
Add referrals table v2

kelvin-chappell:


kelvin-chappell:
This is to take data written in https://github.com/guardian/ophan-data-lake/pull/1104

cloudformation/dynamoDb.yaml
```diff
@@ -220,6 +220,30 @@ Resources:
           - referrals
     DeletionPolicy: Retain
 
+  CampaignReferralsV2DBTable:
```
LATaylor-guardian, line 223: Is there a way we can name this so it's agnostic of the temporary api versioning we put in place (as I plan to remove this at a later date)?

kelvin-chappell: This isn't linked to the api versioning.  It's just that the data structure has changed and I don't want to break the app.  Will remove versioned table names when data is stable and just use original referrals table.

LATaylor-guardian: Got it. In that case 👍

kelvin-chappell:

PR #130, opened 2017-08-30
Add median attention time to campaign performance overview

janua:
## Changes

This adds a new field to the campaign performance overview page; median attention time.

It adds both median attention time over all platforms, and also breaks it down per platform, which can be displayed if you hover over the text.

## Screenshots

nohover

onhover

@LATaylor-guardian

### DIFF
--- a/public/components/CampaignPerformanceOverview/CampaignPerformanceOverview.js
+++ b/public/components/CampaignPerformanceOverview/CampaignPerformanceOverview.js
@@ -1,4 +1,25 @@
 import React, { PropTypes } from 'react';
+import R from 'ramda';
+
+class AttentionTimePerPlatform extends React.Component {
+  render () {
+    const items = Object.keys(this.props.medianAttentionTimeSeconds).map ( (platform) => {
+        return (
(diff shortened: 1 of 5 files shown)


public/components/CampaignPerformanceOverview/CampaignPerformanceOverview.js
```diff
@@ -1,4 +1,25 @@
 import React, { PropTypes } from 'react';
+import R from 'ramda';
```
LATaylor-guardian, line 2: How is Ramda used here?

LATaylor-guardian: Oh I see.

LATaylor-guardian:
Looks good nice work 👍

public/components/CampaignPerformanceOverview/CampaignPerformanceOverview.js
janua: `R.isEmpty` to check if an object is empty.

PR #133, opened 2017-09-05
Median attention time device

janua:
Switch the popup under `median attention time` to use `device_type` instead of `platform`. We also normalise the data instead of leaving as the raw `device_type` data.


LATaylor-guardian:
👍

PR #135, opened 2017-09-07
Show full dataset of referrals per campaign

kelvin-chappell:
This now shows results for all platforms, so that the sum of all the referral clicks should equal the number of page views for the campaign.

### DIFF
--- a/app/repositories/CampaignReferralRepository.scala
+++ b/app/repositories/CampaignReferralRepository.scala
@@ -2,14 +2,10 @@ package repositories
 
 import java.time.LocalDate
 
-import com.amazonaws.services.dynamodbv2.document.Item
-import model.{CampaignReferral, Component}
-import play.api.Logger
-import play.api.libs.json._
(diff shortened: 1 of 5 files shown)


kelvin-chappell:

app/repositories/CampaignReferralRepository.scala
```diff
@@ -1,17 +1,85 @@
...
+  path: Option[String],
+  containerIndex: Option[Int],
+  containerName: Option[String],
+  cardIndex: Option[Int],
+  cardName: Option[String],
+  clicks: Long,
+  unparsedComponent: Option[String]
+)
+
+object CampaignReferralRow {
```
LATaylor-guardian: Can we move this under `models` in its own file? I think this is how the rest of the domain models are defined in this application.

LATaylor-guardian:
👍

janua:
:+1:

PR #136, opened 2017-09-07
Update import mechanism to properly set state upon failure

janua:
I can't see why it would fail in setting the state to `{importing: false}` unless an exception in the `Promise` has occurred.

This updates to catch all exceptions and if a response exists, show it to the user.

### DIFF
--- a/public/components/CapiImport/CapiImport.js
+++ b/public/components/CapiImport/CapiImport.js
@@ -77,11 +77,7 @@ class CapiImport extends Component {
         this.context.router.push('/campaign/' + campaign.id);
       }).catch((error) => {
         console.error(error);
-        if (error.response) {
-          this.setState({importing: false, error: error.response});
-        } else {
-          this.setState({importing: false});
-        }
+        this.setState({importing: false, error: 'Something went wrong; please try again'});
       });
 
     } else {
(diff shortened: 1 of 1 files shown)


kelvin-chappell (approved):
Looks good to me!

public/components/CapiImport/CapiImport.js
```diff
@@ -75,14 +75,21 @@ class CapiImport extends Component {
 
         this.setState({importing: false});
         this.context.router.push('/campaign/' + campaign.id);
-      })
+      }).catch((error) => {
+        console.error(error);
+        if (error.response) {
+          this.setState({importing: false, error: error.response});
```
LATaylor-guardian: Is error.response here a message that is user friendly?

LATaylor-guardian:
👍

public/components/CapiImport/CapiImport.js
janua: Probably not. Shall I remove it? The message I saw was; `campaign does not exist`. But if it was something else it could be actually something not good.

LATaylor-guardian: Maybe just give the user a message that they should try again as an issue occurred.

PR #137, opened 2017-09-07
Bump dependency versions

kelvin-chappell:
I wanted to upgrade to SBT 1.0 but Play isn't ready for that yet!
So instead I've bumped all the dependencies to their latest versions and updated the deprecated code.


janua:
:+1:
That `var` though... 😄

kelvin-chappell:
Not my work!  I'll have a look at it.  Not sure why it can't just be lazy.

PR #138, opened 2017-09-08
Modernise to AMIgo and deb packages

sihil:
Update the CFN, riff-raff.yaml and build process to:
 - create a debian package and then download and install this on instance boot
 - use images from AMIgo that are updates automatically when deployed
 - use coursier for faster build times

I had forgotten that this wasn't really looked after by Ed Tools anymore so here is some free update work for y'all.

__*Note that the CFN change and a new deploy must be done in lockstep. Any outstanding PRs must rebase after this in order to be able to build and deploy.*__

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -35,14 +35,28 @@ lazy val dependencies = Seq(
   "org.scalatest" %% "scalatest" % "3.0.3" % Test
 )
 
-lazy val root = (project in file(".")).enablePlugins(PlayScala, SbtWeb, RiffRaffArtifact)
+lazy val root = (project in file(".")).enablePlugins(PlayScala, SbtWeb, RiffRaffArtifact, JDebPackaging, SystemdPlugin)
   .settings(Defaults.coreDefaultSettings: _*)
   .settings(
+    javaOptions in Universal ++= Seq(
+      "-Dpidfile.path=/dev/null",
+      "-J-XX:MaxRAMFraction=2",
+      "-J-XX:InitialRAMFraction=2",
+      "-J-XX:MaxMetaspaceSize=500m",
+      "-J-XX:+UseConcMarkSweepGC",
+      "-J-XX:+PrintGCDetails",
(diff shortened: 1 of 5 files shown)


sihil:
[Campaign Central: Update AMI / Riff-Raff](https://trello.com/c/0YwAxpV2/187-campaign-central-update-ami-riff-raff)

build.sbt
```diff
@@ -35,14 +35,28 @@ lazy val dependencies = Seq(
...
+      "-J-XX:MaxRAMFraction=2",
+      "-J-XX:InitialRAMFraction=2",
+      "-J-XX:MaxMetaspaceSize=500m",
+      "-J-XX:+UseConcMarkSweepGC",
+      "-J-XX:+PrintGCDetails",
+      "-J-XX:+PrintGCDateStamps",
+      s"-J-Xloggc:/var/log/${packageName.value}/gc.log"
+    ),
+    debianPackageDependencies := Seq("openjdk-8-jre-headless"),
+    maintainer := "Editorial Tools Developers <digitalcms.dev@theguardian.com>",
```
kelvin-chappell: Might want to change this to `Commercial Dev Team <commercial.dev@theguardian.com>`

cloudformation/campaign-central.yaml
```diff
@@ -156,11 +151,11 @@ Resources:
           Action:
           - s3:GetObject
           Resource:
-          - arn:aws:s3:::github-team-keys/*
+          - arn:aws:s3:::github-public-keys/*
```
kelvin-chappell, line 154: Does this exist?

kelvin-chappell (approved):
Thanks for this! 👍

cloudformation/campaign-central.yaml
sihil: It's the new 'supported' version that is in the central deploy tools account rather than the EdTools account. AMIgo should be using the new location and I've just checked that your image is indeed.

PR #140, opened 2017-09-11
Add dwell time data per campaign.

LATaylor-guardian:
Provides the weighted average dwell time per campaign along with the average dwell time per path.

picture 140


janua:
Some things will have a massive popup though? Can probably fix it when it comes about though...

👍

LATaylor-guardian:
> Some things will have a massive popup though? Can probably fix it when it comes about though...

Yeah definitely. I've just stuck it there for now for want of a better way of displaying it.

PR #141, opened 2017-09-12
Fix import form to carry values

janua:
This updates the import form;
 - Carry `campaignValue`, `uniquesTarget` and `pageviewsTarget` into the campaign
 - Fix the workflow so that it brings you to the campaign you have just created/already exists
 - Remove autocomplete from the form
 - Fix the validation on the form to show errors on whole form rather than only on the field being edited

### DIFF
--- a/app/model/command/ImportCampaignFromCAPICommand.scala
+++ b/app/model/command/ImportCampaignFromCAPICommand.scala
@@ -142,9 +142,7 @@ case class ImportCampaignFromCAPICommand(
         case Some(_) => "paidContent"
         case None => InvalidCampaignTagType
       }
-
-      Some("uniques" -> uniquesTarget) ++ pageviewTarget.map("pageviews" -> _).toMap
-
+      
       val campaign = CampaignRepository.getCampaignByTag(tag.id) getOrElse {
         Campaign(
           id = UUID.randomUUID().toString,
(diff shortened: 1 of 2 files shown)


app/model/command/ImportCampaignFromCAPICommand.scala
```diff
@@ -120,38 +127,38 @@ case class ImportCampaignFromCAPICommand(
...
+          pathPrefix = Some(tag.section.pathPrefix),
           clientId = findOrCreateClient(sponsorship).id,
           created = now,
           createdBy = userOrDefault,
           lastModified = now,
           lastModifiedBy = userOrDefault,
-          nominalValue = Some(10000),         // default targets and values
-          actualValue = Some(10000),          // these will be set in the UI manually
-          targets = Map("uniques" -> 10000L)
+          nominalValue = None,                        // default targets and values
```
janua, line 159: @kelvin-chappell Set this to `None` as the default; I can't see any references to it anywhere other than here.

LATaylor-guardian:
👍

app/model/command/ImportCampaignFromCAPICommand.scala
kelvin-chappell: That nominal value looks like a nominal uniques target to me.  So it is a bit useless.

PR #142, opened 2017-09-12
Replace pandomain auth with google auth

kelvin-chappell:
This gives access to all endpoints to anyone with a gu email address.

There isn't an extra level of authorisation for update actions, as changes are audited.  If it becomes necessary we can require group membership but there isn't an awful lot you can update at the moment so it seems like overkill.


app/controllers/LogIn.scala
```diff
@@ -0,0 +1,32 @@
+package controllers
+
+import com.gu.googleauth.{GoogleAuthConfig, LoginSupport}
+import play.api.libs.ws.WSClient
+import play.api.mvc.{BaseController, Call, ControllerComponents}
+
+import scala.concurrent.ExecutionContext
+
+// based on https://github.com/guardian/play-googleauth/blob/master/example/app/controllers/Login.scala
+class LogIn(
```
LATaylor-guardian, line 10: Small point - should be `class Login`

app/controllers/LogIn.scala
```diff
@@ -0,0 +1,32 @@
...
+  override val wsClient: WSClient,
+  val controllerComponents: ControllerComponents,
+  val authConfig: GoogleAuthConfig
+)(implicit executionContext: ExecutionContext)
+  extends LoginSupport
+  with BaseController {
+
+  def logIn() = Action.async { implicit request =>
```
LATaylor-guardian, line 18: Same here `def Login`

app/controllers/LogIn.scala
kelvin-chappell: 'Log in' is two words, isn't it? A noun phrase and a verb phrase.  There's no such thing as a login!

LATaylor-guardian:
I think it's probably worth adding `Scalariform` to the project to make the formatting of code less opinionated. 

If you add `addSbtPlugin("org.scalariform" % "sbt-scalariform" % "1.8.0")` to plugins.sbt that should do it.

app/controllers/LogIn.scala
LATaylor-guardian: It's widely used across the internet in this context as `login` no? https://www.google.co.uk/search?q=login&oq=login&aqs=chrome..69i57j69i60l4j0.1036j0j4&sourceid=chrome&ie=UTF-8

LATaylor-guardian:
After adding Scalariform 👍

kelvin-chappell:
That's a good idea.  I'll add it.  I think I'd better do it in a separate PR.

philwills (approved):
This looks good to me.

I agree about doing the auto-format in a separate PR. http://scalameta.org/scalafmt/ is shinier, so obviously has my  vote for that.

app/controllers/LogIn.scala
kelvin-chappell: In that case I think it could be either, so I'll merge it as-is but feel free to change it later.

kelvin-chappell:
I've been using scalafmt in Intellij so I'll use that as I'm familiar with it.

LATaylor-guardian:
@kelvin-chappell Does `scalafmt` require zero configuration like `scalariform`?

kelvin-chappell:
Yes

PR #145, opened 2017-09-12
Format scala files with scalafmt

kelvin-chappell:
Set up scalafmt to run before compile.
 
The key commit is 1e892a5fa5bc4f4f7d62ecbce250bf2a919e5a53


LATaylor-guardian:
I'm not a huge fan of `defaultWithAlign` but this is largely a matter of opinion so 👍

PR #148, opened 2017-09-13
Refactor API

janua:
Generally speaking, this PR moves the API in a direction where we model most states reliably in order to react to them or at the very least log them.

### This PR does a few(!) things:
 - Remove the unnecessary `trait Command` trait
 - Start adhering to the `Either` type `process` is meant to return; this changes both `ImportCampaignFromCAPICommand` and `RefreshCampaignFromCAPICommand`
 - Start moving away from using `Exception`s as a control flow mechanism
 - Start using more referentially transparent methods; stop using side affects
 - Make code dependencies in methods actually depend on each other (!)
 - Model most of the common problems via `CampaignCentralApiError`

### Things that could/should be done but are not yet done;
 - Remove `CommandError` and all it's exception friends and usages
 - Remove `CAPIImportCommand` as a mixin
 - Properly return appropriate HTTP codes to the client and have the client handle the scenarios properly; if a campaign already exists, the user is taken to the campaign rather than being told it already exists
 - Upgrade to scala 2.12 to get the right biased `Either`
 - Remove the way `process` works, the process method depends on an objects instantiation properties; are we back to OO again?
 - Perhaps reduce the amount of steps to proxy through just to do something?

### Other bigger things I have noticed
 - A synchronous API, really? How about `DynamoDbAsyncClient`? In my opinion, you do not need the `Either` API if you were using `Futures`, although it is still useful.
 - We seriously need to fix `ContentApi.loadAllContentInSection`; I can't see why it can't take a method that lazily asserts whether it can derive the tag as the content passes through, rather than getting ALL the content only to then derive


app/model/command/ImportCampaignFromCAPICommand.scala
```diff
@@ -29,7 +29,10 @@ trait CAPIImportCommand extends Command {
   }
 
   def deriveContentType(apiContent: ApiContent) = {
-    apiContent.tags.find(_.`type` == TagType.Type).map(_.webTitle).getOrElse(UnableToDetermineContentType)
+    apiContent.tags
+      .find(_.`type` == TagType.Type)
+      .map(_.webTitle)
+      .getOrElse(throw new RuntimeException(s"Unable to derive content type from ${apiContent.id}"))
```
LATaylor-guardian: Why are you throwing a runtime exception here?

LATaylor-guardian:
Nice work 👍

app/model/command/ImportCampaignFromCAPICommand.scala
janua: Because I haven't removed all the exceptions yet. Actually, this was meant to replace the removal of `CommandError` but I put `CommandError` back in as there were too many usages, but didn't revert this!

PR #150, opened 2017-09-14
Remove the default User

janua:
There is no need to wrap `User` in an `Option` only to the try to take it back out and then fail back to the `defaultUser` if it doesn't exist; it should always exists.

This means all process now implicitly require a `User` object in scope.


kelvin-chappell (approved):
👍  but I wonder why it was optional

janua:
@kelvin-chappell No idea, I reckon something at the start made it Optional. Maybe it could come out at `null` or something? It's not possible for it to come out as `null` now though!

PR #151, opened 2017-09-14
Refactor command

janua:
This moves away from using case class instances to process the requests into plain methods that take in commands they work on;

Changes include;
 - Move all commands to a `Commands` object
 - Rename `process` to an appropriate name of what it will do
 - Removal of unnecessary case class `RefreshCampaignFromCAPICommand`


kelvin-chappell (approved):
Seems sensible 👍

PR #155, opened 2017-09-18
A better readme

shtukas:


README.md
```diff
@@ -1,50 +1,49 @@
 Campaign Central
 ================
 
- 
-Running
-=======
+Install and Setup
+=================
```
kelvin-chappell, line 5: Might be better to use '##' for level 2 heading level so that it's consistent with the level 3s.

janua:
👍 
Maybe we should just encourage everyone to use the start bash script; I don't know why you'd ever want to run `sbt run` by itself! Unless we remove the javascript

README.md
```diff
@@ -1,50 +1,49 @@
...
+include sites-enabled/*;
+server_names_hash_bucket_size 256;
+```
+
+### configuration
 
-### nginx Troubleshooting
-If you get the following error when attempting to start nginx after adding campaign central config:
+Before you run for the first time you will need to run `scripts/setup.sh`.  This will install and compile all the frontend dependencies needed for the app (you may need to install npm before you can run this successfully). If any frontend dependencies are changed you should should re-run the setup script.
+
```
kelvin-chappell, line 21: should should

shtukas:
Yeah, I noticed that. Actually we do in the section "Client Side Development". But I recon the text is not perfect. Will do for now...

README.md
```diff
@@ -1,50 +1,49 @@
...
+You will then need to create a local configuration file. The easiest way to do this is to copy the example config:  
 
 ```
-could not build the server_names_hash, you should increase server_names_hash_bucket_size: 64
+cp conf/example-local.conf conf/local.conf
 ```
 
-Add the following to your `/usr/local/etc/nginx/nginx.conf` in the `http { }` block:
+The campaign-central is a standard play app, so to run it's `sbt run`.  You will need to have developer credentials configured in an AWS composer profile to be able to run it.  To test that the play app is running independently of nginx, try [hitting http://localhost:2267](hitting http://localhost:2267).
```
kelvin-chappell, line 29: <del>The</del>

kelvin-chappell (approved):
Good 👍

PR #158, opened 2017-09-18
Fix the z-index of the popover

janua:
Sometimes the dwell time popover ends up being long and going under the chart beneath.


LATaylor-guardian:
👍

PR #160, opened 2017-09-19
Small improvements.

LATaylor-guardian:
- Remove unneeded code.
 - Make header banner text shorter so it doesn't line break.


shtukas:
:+1:

PR #162, opened 2017-09-20
Update dependencies

kelvin-chappell:
Bumped some, removed others and updated deprecated code.


janua:
Some sneaky CSS changes too? 😄 
👍

kelvin-chappell:
I didn't do those!  That was the setup script.  But I assume it does no harm to check them in.

LATaylor-guardian:
Those CSS changes need to be reverted. This happened to me locally and broke all of the icons.

PR #164, opened 2017-09-20
Add a Glossary

janua:
This adds a glossary page that describes the terms; it also adds a question mark beside the terms in the campaign performance metrics page that when you hover over, will display the text.

@LATaylor-guardian Is working on something similar that reworks the metrics page; so this question mark piece may get removed.

@kelvin-chappell

### DIFF
--- a/public/components/Glossary/Glossary.js
+++ b/public/components/Glossary/Glossary.js
@@ -1,6 +1,6 @@
 import React from 'react';
 
-const MedianAttentionTimeExplainerText = 'Attention time is the measure of how long a user is knowingly active on the page; they have recently moved their mouse or scrolled.';
+const MedianAttentionTimeExplainerText = 'Attention time is the measure of how long a user is knowingly active on the page; they have recently moved their mouse or scrolled within 5 seconds, or are watching a video in a focused tab.';
 const DwellTimeExplainerText = 'Dwell time tries to emulate the same viewability metric from Google Analytics. It is calculated by taking the difference between a pageview timestamp and the subsequent pageview timestamp. These pageviews have to happen within a 30 minute session period, and are averaged when there has been more than one pageview for a unique user.';
 
 class Glossary extends React.Component {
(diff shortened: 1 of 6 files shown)


public/components/Glossary/Glossary.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+const MedianAttentionTimeExplainerText = 'Attention time is the value where a user is knowingly active on the page; they have recently moved their mouse or scrolled.';
```
kelvin-chappell: Perhaps `... is the measure of how long a user ...` ?

kelvin-chappell: Is it worth saying how long recent is, or is that deliberately left unstated?

public/components/Glossary/Glossary.js
```diff
@@ -0,0 +1,20 @@
...
+const DwellTimeExplainerText = 'Dwell time tries to emulate the same viewability metric from Google Analytics. It is calculated by taking the difference between a pageview timestamp and the subsequent pageview timestamp. These pageviews have to happen within a 30 minute session period, and are averaged when there has been more than one pageview for a unique user.';
```
kelvin-chappell: Perhaps `as Google` instead of `from Google`?

kelvin-chappell (approved):
Just some minor pedantic points
👍

public/components/Glossary/Glossary.js
janua: I think `from` makes more sense? I don't understand exactly what `as` would mean! I think it might not even be called dwell time in GA either, it might be time on page... Need to check.

public/components/Glossary/Glossary.js
janua: Yes, that sounds better...
The time is left deliberately unstated as I don't know the exact time you need to be inactive for

janua:
@kelvin-chappell Seems to be [5 seconds](https://github.com/guardian/ophan/blob/master/tracker-js/assets/coffee/ophan/attention.coffee); how is that for wording now?

kelvin-chappell:
🥇

PR #165, opened 2017-09-20
Improve display of metrics.

LATaylor-guardian:
This PR provides  react components that make it more simple and reusable to display metrics and their associated breakdown data. This PR also attempts to begin to structure and display data in a more user friendly way.

Before:
picture 149

After: 
picture 147

When you click on + icon in top right hand corner:
picture 148


kelvin-chappell (approved):
Nice! 👍

PR #168, opened 2017-09-22
Stop breaking down referrals by edition

kelvin-chappell:
One piece of feedback was that breaking down referrals by edition wasn't very useful.
So I've taken it out.


janua:
:+1:

LATaylor-guardian:
👍

PR #169, opened 2017-10-03
Improve refreshCampaigns endpoint to import and update campaigns automatically.

LATaylor-guardian:


app/repositories/contentapi/CapiSectionTransformer.scala
```diff
@@ -0,0 +1,36 @@
...
+    section.activeSponsorships.flatMap(_.headOption.map(_.sponsorshipType.name))
+
+  def deriveSponsorshipLogo(section: CapiSection): Option[String] =
+    section.activeSponsorships.getOrElse(Nil).headOption.map(sponsorship => sponsorship.sponsorLogo)
+
+  def deriveStatus(startDate: Option[DateTime], endDate: DateTime): String = {
+    (startDate, endDate) match {
+      case (_, ed) if ed.isBeforeNow => "dead"
+      case (Some(sd), _)             => "live"
+      case _                         => "unknown"
```
kelvin-chappell: In this case there's no start date and end date is in future, which would make it `live`.

app/services/CampaignService.scala
```diff
@@ -97,4 +94,103 @@ object CampaignService {
...
+    val contentItems: List[ContentItem] = CapiContentTransformer.buildContentItems(apiContent, campaign.id)
+
+    val putContentResults: List[Either[CampaignCentralApiError, PutContentItemResult]] = for {
+      item <- contentItems
+    } yield CampaignContentRepository.putContent(item)
+
+    val results: (List[CampaignCentralApiError], List[PutContentItemResult]) =
+      putContentResults.foldRight((List[CampaignCentralApiError](), List[PutContentItemResult]())) {
+        case (itemResultOrError, (ls, rs)) => itemResultOrError.fold(l => (l :: ls, rs), r => (ls, r :: rs))
+      }
```
kelvin-chappell: If only interested in first error, could do:
```
putContentResults.collectFirst {
      case Left(e) => e
    } toLeft {
      putContentResults collect {
        case Right(result) => result
      }
    }
```

kelvin-chappell (approved):
Looks good! 👍 

Not sure what triggers the refresh.

LATaylor-guardian:
@kelvin-chappell 
>Not sure what triggers the refresh.

I believe the lambda swells wrote does: https://github.com/guardian/campaign-central-schedule-lambdas/blob/master/functions/expireCampaigns/main.go#L64

app/services/CampaignService.scala
LATaylor-guardian: Thanks, I just moved this code from another file but will change as much clearer.

app/repositories/contentapi/CapiSectionTransformer.scala
LATaylor-guardian: Thanks will update.

PR #172, opened 2017-10-05
We should use web title of section for campaign name rather than sponsor name.

LATaylor-guardian:


app/services/CampaignTransformer.scala
```diff
@@ -18,7 +18,7 @@ object CampaignTransformer {
     val endDate   = CapiSectionTransformer.deriveEndDateOrDefaultToNow(section)
 
     for {
-      sponsorName     <- CapiSectionTransformer.deriveSponsorName(section)
+      sponsorName     <- Some(CapiSectionTransformer.deriveCampaignName(section))
```
janua: Why not make it a `val` now?

janua:
:+1:

app/services/CampaignTransformer.scala
LATaylor-guardian: God knows - wasn't concentrating. Have changed.

PR #175, opened 2017-10-06
Create Lambda function to invoke the refreshing of campaigns.

LATaylor-guardian:
I have purposely not done the following so that it can be left as a learning exercise for someone who hasn't done this type of thing before:

- Make lambda deployable via Riff Raff.
- Cloudform lambda creation.
- Move the code responsible for refreshing the campaigns into the lambda.
- Consider and implement a means for knowing when this Lambda fails and alerting the team.
- build on Team City.


janua:
:+1:

PR #179, opened 2017-10-09
Add method to return the desired string for referral path

katebee:
#### What is changing and why:

fix for the path coming from native apps within the referrals table ([trello card](https://trello.com/c/Wlw8n5LK/111-campaign-central-referrals-fix-path-coming-from-native-apps))

Android and iOS apps are currently including the edition as a prefix on the path, for example: `uk/royal-canin-cats` should actually be: `/royal-canin-cats`

This adds a method that deals with the gettingOrElse of path, and strips any `uk` prefix

##### Before:

picture 38

##### After:

picture 36

### DIFF
--- a/app/repositories/CampaignReferralRepository.scala
+++ b/app/repositories/CampaignReferralRepository.scala
@@ -29,7 +29,7 @@ object CampaignReferralRepository {
             CampaignReferral(
               component = Component(
                 platform = formatPlatform(platform),
-                path = path getOrElse "unknown",
+                path = formatPath(path),
                 containerIndex = containerIndex getOrElse 0,
                 containerName = groupedValues.headOption.flatMap(_.containerName) getOrElse "",
                 cardIndex = cardIndex getOrElse 0,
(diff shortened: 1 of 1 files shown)


shtukas (approved):

app/repositories/CampaignReferralRepository.scala
```diff
@@ -60,4 +60,12 @@ object CampaignReferralRepository {
     case "WINDOWS_NATIVE_APP" => "Windows"
     case other                => other
   }
+
+  private def formatPath(maybePath: Option[String]): String = {
+    maybePath match {
+      case Some(s) if s.startsWith("uk/") => s.stripPrefix("uk")
```
kelvin-chappell, line 66: I think app paths can be prefixed by any edition name, so probably need a case for `us`, `au` and whatever the name of the international edition is, possibly `international`.

kelvin-chappell (commented):
Thanks for looking at this.  At the moment there's also some inconsistency in whether paths start with `/` or not.  They should probably all start with `/` as you've got here, but that's another issue.

PR #180, opened 2017-10-11
Provide initial breakdown of campaign performance by location and device.

LATaylor-guardian:
Still to do in seperate PR:
 - Dwell time by device.
 - Dwell time by country code.
 - Median attention time by device.
 - Median attention time by country code.
 - Dwell time/median attention time/page views/uniques by path.

Screenshots:
-------------
By country:
picture 151

by device: 
picture 152


LATaylor-guardian:
@kelvin-chappell If you've had a chance to look over Chloe's slides, I think we should add referrals as a button and then show the referrals view you have done in the same place as the above. This would tie into some familiar view for them from GA. What do you reckon?

PR #184, opened 2017-10-13
Add benchmarks to Campaign Central.

LATaylor-guardian:
Benchmarks is a new page, requested by Chloe, that looks to provide the metrics we provide at campaign level across ALL campaigns. This at current is done manually in a spreadsheet and repeated by one or more people. This information is generally used to give an idea of what sales can use to sell and is informative in that process.
picture 155

### DIFF
--- a/app/model/Benchmarks.scala
+++ b/app/model/Benchmarks.scala
@@ -1,0 +1,47 @@
+package model
+
+import ai.x.play.json.Jsonx
+import play.api.libs.json.Format
+import util.DoubleUtils._
+
+case class Totals(uniques: Long, pageviews: Long, timeSpentOnPage: Double)
+object Totals {
+  implicit val totalsFormat: Format[Totals] = Jsonx.formatCaseClass[Totals]
+
+  def apply(latestAnalytics: Seq[LatestCampaignAnalytics]): Totals = {
+    Totals(
+      uniques = latestAnalytics.map(_.uniques).sum,
+      pageviews = latestAnalytics.map(_.pageviews).sum,
+      timeSpentOnPage = latestAnalytics.flatMap(_.weightedAverageDwellTimeForCampaign).sum.to2Dp
+    )
+  }
+}
+
+case class Averages(uniques: Long, pageviews: Long, timeSpentOnPage: Double)
+object Averages {
+  implicit val averagesFormat: Format[Averages] = Jsonx.formatCaseClass[Averages]
+
+  def apply(latestAnalytics: Seq[LatestCampaignAnalytics]): Averages = {
+    val noOfAnalytics = latestAnalytics.size.toLong
+    Averages(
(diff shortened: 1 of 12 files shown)


shtukas:
+1 LGTM

app/model/Benchmarks.scala
```diff
@@ -0,0 +1,47 @@
+package model
+
+import ai.x.play.json.Jsonx
+import play.api.libs.json.Format
+import util.DoubleUtils._
+
+case class Totals(uniques: Long, pageviews: Long, timeSpentOnPage: Double)
+object Totals {
+  implicit val totalsFormat: Format[Totals] = Jsonx.formatCaseClass[Totals]
```
JonNorman, line 9: How come you aren't just using the inbuilt Play `Format` here?

app/services/CampaignService.scala
```diff
@@ -61,6 +61,34 @@ object CampaignService {
     }
   }
 
+  def getBenchmarksAcrossCampaigns(): Either[CampaignCentralApiError, Benchmarks] = {
```
JonNorman, line 64: How come you use `Either`s and not `scala.util`'s `Success`/`Failure`?

conf/routes
```diff
@@ -6,6 +6,7 @@ GET	/campaign/:id		              controllers.App.index(id)
 GET	/capiImport		                controllers.App.index(id = "")
 GET	/management/analytics         controllers.App.index(id = "")
 GET	/glossary                     controllers.App.index(id = "")
+GET	/benchmarks                   controllers.App.index(id = "")
```
JonNorman, line 9: Is there a reason that the routes are set up like this (with `id` set to the empty string?). Looks like the `id` should be optionally part of the request path or just parsed from the path within `controllers.App.index` handler.

JonNorman (commented):
Looks good, just a few questions from me (for my own edification).

app/services/CampaignService.scala
LATaylor-guardian: I typically avoid Try in Scala unless I'm using Java libraries that may throw exceptions. Even then I will convert from Try and model my exceptions using an ADT and typically return an Either. This has become much nicer now in 2.12 given Either is right biased.

conf/routes
LATaylor-guardian: Yes. When your in a single page application and using react-router you never make a request away from the page to the server. But if you were on say `/foo` and you refreshed it would hit the server and at that point Play's router. At this point you want the server to send you back to the single page application and back into the client side router so the right page is shown to the user.

app/model/Benchmarks.scala
LATaylor-guardian: It was already in the project. It's essentially play json still just suked up with some extras: https://github.com/xdotai/play-json-extensions

PR #185, opened 2017-10-13
Show referral impressions and CTRs

kelvin-chappell:
This adds two extra columns to the referrals table:
image


app/model/CampaignReferral.scala
```diff
@@ -20,11 +20,29 @@ object Component {
 // A referral to any item in a campaign from an on-platform source in a particular period
 case class CampaignReferral(
   component: Component,
-  numClicks: Int,
+  clickCount: Int,
+  impressionCount: Int,
+  // Annoyingly react won't recognise CTR if it's a derived property so it has to go in as part of the constructor
```
LATaylor-guardian: I don't understand what you mean here, can you explain please?

LATaylor-guardian:
👍

app/model/CampaignReferral.scala
kelvin-chappell: Thanks for questioning it.  I realise that it makes no sense at all, and I've fixed the underlying problem now.

LATaylor-guardian: Great :)

PR #186, opened 2017-10-16
Remove unused modal from big card metrics.

LATaylor-guardian:
We now display this data in a component underneath in a view more familiar to that of GA.


janua:
:+1:

PR #187, opened 2017-10-16
Only show top 20 countries.

LATaylor-guardian:
Chloe only cares about the top 20 for now rather the 200 odd showing now and cluttering the view.


janua:
What about the tail? Maybe add them together if you wanted to keep them?
:+1:

LATaylor-guardian:
> What about the tail? Maybe add them together if you wanted to keep them?

@janua Add them together and display as other do you mean?

janua:
Yep! Exactly, nice! 👍

PR #188, opened 2017-10-16
Ensure Hosted campaigns are marked as so.

LATaylor-guardian:
A hosted campaigns ID will always begin with 'advertiser-content' as opposed to paid for type campaigns which do not.


janua:
:+1:

kelvin-chappell (approved):
👍  It looks like a safe assumption

PR #197, opened 2017-10-19
Tidy header

rich-nguyen:
Tidies the header, by making the it responsive and removing the sticky positioning. Only the sidebar is sticky, for a cleaner view.

***Before***
screen shot 2017-10-19 at 13 05 00

***After***
screen shot 2017-10-19 at 13 24 07


LATaylor-guardian:
👍  🖌 🎨

PR #198, opened 2017-10-19
Fix campaign layout to fill full width of screen.

LATaylor-guardian:


LATaylor-guardian:
@rich-nguyen Thanks for the explanation.

public/styles/components/main/_main.scss
```diff
@@ -5,6 +5,7 @@
 
 .main__content {
     padding: 40px;
+    flex:1;
```
rich-nguyen: might be safer to `1 0 auto` for IE

rich-nguyen:
👍

PR #200, opened 2017-10-24
Share content width for benchmarks

rich-nguyen:
The benchmarks page also needs a width set, so do so on main content


scottp719 (approved):
:+1:

PR #201, opened 2017-10-25
Upgrade capi client (which contains a change in http client)

LATaylor-guardian:
This is required so that I can make use of the new `sponsorship-type` filter I have added to the `Tags` endpoint.


janua:
:+1:

PR #203, opened 2017-10-26
Add median attention time for campaign benchmarks.

LATaylor-guardian:


app/model/Benchmarks.scala
```diff
@@ -31,16 +31,34 @@ object Averages {
...
+  implicit val mediansFormat: Format[Medians] = Jsonx.formatCaseClass[Medians]
+
+  def apply(latestAnalytics: Seq[LatestCampaignAnalytics]): Medians = {
+    Medians(
+      attentionTime = {
+        val attnTimes      = latestAnalytics.flatMap(_.medianAttentionTimeSeconds)
+        val (lower, upper) = attnTimes.sorted.splitAt(attnTimes.size / 2)
+
+        if (attnTimes.size % 2 == 0) {
+          (lower.lastOption.map(_.toDouble).getOrElse(0.0) + upper.headOption.map(_.toDouble).getOrElse(0.0)) / 2.0
```
janua, line 45: Are you meant to do this? Get the average between the pivot values?

LATaylor-guardian: @janua I believe so if the number of items in your dataset is even unless I am mistaken?

janua:
:+1:

PR #205, opened 2017-10-31
Blacklist sections that we do not want to create campaigns from.

LATaylor-guardian:
Tim Hughes can confirmed that we do not require these campaigns.


JonNorman (approved):
Are you doing anything else within the object? Why not just have a `blacklistedSections: List[String]`?

LATaylor-guardian:
> Are you doing anything else within the object? Why not just have a blacklistedSections: List[String]?

Yeah good point, I got caught between doing two things and didn't change back. Will change.

PR #206, opened 2017-10-31
Move header to play template

kelvin-chappell:
This is the first step to get rid of the redux, in an attempt to improve (my) productivity.


LATaylor-guardian:
This isn't making any changes to Redux(state management: https://github.com/reactjs/redux), but rather moving from rendering the header client side with React to rendering it server side with Play templates. 

@kelvin-chappell I'm happy to spend some time helping you get up to speed with React/Redux if you feel this would be helpful? I'm by no means an expert but can navigate it.

As an aside, this is probably worth us all discussing together (those working on Campaign Central @janua @rich-nguyen) whether such a substantial refactoring would be the best investment of our time.

kelvin-chappell:
It would be interesting to work out the best approach to convert a react/redux project to a conventional play app, but it is a lot of work and difficult to do it in small chunks.

PR #212, opened 2017-11-03
Media events component

rich-nguyen:
This adds a table to the campaign page for any media events that were obtained from the db.

screen shot 2017-11-03 at 11 25 17

I could tidy up the path names later on. This can go live, but there won't be any data until `commercial_pageview` is live.


LATaylor-guardian:
Looks great! Nice work @rich-nguyen 👍

PR #213, opened 2017-11-03
Add view-through rate column for Media Events tables

rich-nguyen:
Add a calculated column for media events tables:

screen shot 2017-11-03 at 14 33 05


LATaylor-guardian:
Lovely job 👍

PR #215, opened 2017-11-06
Present on-platform referrals as a tree structure

kelvin-chappell:
Instead of a flat structure, the on-platform referrals table is now arranged as a tree with these levels:
container > card > platform > individual dates.

Eg.
Initial collapsed state:
image

Fully-expanded state:
image


janua:
:+1:

PR #218, opened 2017-11-07
Merge referring containers by name

kelvin-chappell:
The index of the container isn't important.

Before:
image

After:
image

### DIFF
--- a/app/repositories/CampaignReferralRepository.scala
+++ b/app/repositories/CampaignReferralRepository.scala
@@ -41,11 +41,9 @@ object CampaignReferralRepository {
           }
 
         val referrals = subReferrals(filtered)(_.formattedPath)(identity) { pathRows =>
-          Some(subReferrals(pathRows) { row =>
-            (row.containerIndex, row.containerName)
-          } {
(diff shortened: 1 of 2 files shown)


app/repositories/CampaignReferralRepository.scala
```diff
@@ -41,11 +41,9 @@ object CampaignReferralRepository {
...
 
         val referrals = subReferrals(filtered)(_.formattedPath)(identity) { pathRows =>
-          Some(subReferrals(pathRows) { row =>
-            (row.containerIndex, row.containerName)
-          } {
-            case (Some(containerIndex), Some(containerName)) => s"Container #$containerIndex: $containerName"
-            case _                                           => ""
+          Some(subReferrals(pathRows)(_.containerName) {
+            case Some(containerName) => s"Container: $containerName"
+            case _                   => ""
```
janua: `case None =>`

janua:
:+1:

app/repositories/CampaignReferralRepository.scala
kelvin-chappell: Good point

PR #219, opened 2017-11-08
Set root of referrals tree to show global summary for campaign

kelvin-chappell:
Root now says 'All fronts' and gives a single impressions, clicks and CTR value for the whole campaign.

There's a refactor and some of the code has moved around a bit as well.


LATaylor-guardian:
Nice work 👍

PR #225, opened 2017-11-09
Update case class LatestAnalyticsBreakdownItem

shtukas:
To include attributes `facebookShares` and `linkedInShares` which are now expected by the front end.


kelvin-chappell (approved):
👍

PR #229, opened 2017-11-16
Create new table for location specific latest analytics.

LATaylor-guardian:
This will eventually replace the previous table which will be deleted on switchover.

### DIFF
--- a/cloudformation/dynamoDb.yaml
+++ b/cloudformation/dynamoDb.yaml
@@ -148,6 +148,32 @@ Resources:
           - latest
     DeletionPolicy: Retain
 
+  CampaignLocationAnalyticsLatestDBTable:
+    Type: AWS::DynamoDB::Table
+    Properties:
+      AttributeDefinitions:
+      - AttributeName: territory
+        AttributeType: S
+      - AttributeName: campaignId
+        AttributeType: S
+      KeySchema:
+      - AttributeName: territory
+        KeyType: HASH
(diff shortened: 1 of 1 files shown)


cloudformation/dynamoDb.yaml
```diff
@@ -240,6 +240,32 @@ Resources:
...
+    Properties:
+      AttributeDefinitions:
+      - AttributeName: territory
+        AttributeType: S
+      - AttributeName: campaignId
+        AttributeType: S
+      KeySchema:
+      - AttributeName: territory
+        KeyType: HASH
+      - AttributeName: campaignId
```
janua, line 162: Which way around is better, is this going to be displayed inside a particular campaign? If so maybe the opposite might be better?

JonNorman (approved):

cloudformation/dynamoDb.yaml
LATaylor-guardian: @janua I thought about this for a bit. I think this way as Campaign Central will be contextual to the territory which the user will select at the top level of Campaign Central. So this allows you to query for all campaigns in a given territory.

janua: Ah I see... Then you can batch get the campaigns!

LATaylor-guardian: Yeah :)

janua:
:+1:

PR #230, opened 2017-11-17
Set write throughput ratio to 1.0 in dev table copy job

kelvin-chappell:
No other process is writing to dev tables so should be safe.

It seems to take a lot longer than 2 hours to write some of the tables, but the data pipeline job doesn't show an error.  This is why a lot of the campaign data is missing in dev.

By looking at the write metrics for each table we should be able to see how long it actually takes, and change the timeout accordingly.


janua:
:+1:

LATaylor-guardian:
👍

rich-nguyen (approved):

PR #231, opened 2017-11-17
Referrals column ordering

rich-nguyen:
This adds column ordering to the referrals component. I had to re-wire everything.

screen shot 2017-11-17 at 18 49 16


public/components/CampaignAnalytics/Analytics/CampaignReferral.js
```diff
@@ -5,18 +5,21 @@ import "react-infinity-menu/src/infinity-menu.css";
 export default class CampaignReferral extends React.Component {
 
   render() {
+    // The keyPath needs to be halved because it is typically something like:
+    // '0.children.0.children.0.children.0.children.0'
+    // should give a depth of 4.
     const depth = Math.floor(this.props.data.keyPath.split(".").length / 2);
```
kelvin-chappell, line 11: Depth is actually available as an arg in the nodeClick method, but I've no idea how to pass it through.

kelvin-chappell (approved):
👍  but I am concerned that it's quite difficult to maintain.  I wonder if any functionality can be abstracted out but I imagine that would be quite tricky/impossible.

rich-nguyen:
Yeah it looks complicated, but apart from `sortTree()`, most of it is just code moving around. It made sense to move all transformation work into the reducers, so that `CampaignReferrals` operates on `props` rather than `state`. In my first attempt, with the tree left as part of `this.state`, it was awkward dealing with redux for other stuff, like the toggle. Now that redux fully owns all the state, react components in this area just deal with presentation.

PR #233, opened 2017-11-22
Support date-range filtering in referrals fetch

kelvin-chappell:
This is the API change.  The UI change will follow.

### DIFF
--- a/app/controllers/CampaignApi.scala
+++ b/app/controllers/CampaignApi.scala
@@ -1,11 +1,15 @@
 package controllers
 
+import java.time.LocalDate
+
 import com.gu.googleauth.AuthAction
 import model._
 import play.api.mvc._
 import repositories._
 import services.CampaignService
 import play.api.libs.json.Json
+
+import scala.util.Try
 
 class CampaignApi(components: ControllerComponents, authAction: AuthAction[AnyContent])
   extends AbstractController(components) {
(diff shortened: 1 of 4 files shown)


app/controllers/CampaignApi.scala
```diff
@@ -105,8 +109,15 @@ class CampaignApi(components: ControllerComponents, authAction: AuthAction[AnyCo
...
   }
 
-  def getCampaignReferrals(campaignId: String) = authAction { _ =>
-    CampaignReferralRepository.getCampaignReferrals(campaignId) match {
+  def getCampaignReferrals(campaignId: String, start: Option[String], end: Option[String]) = authAction { _ =>
+    def toDate(o: Option[String]): Option[LocalDate] = o flatMap { s =>
+      Try(LocalDate.parse(s)).toOption
+    }
+    val dateRange = for {
+      from <- toDate(start)
```
LATaylor-guardian, line 117: What happens if a date range is not specified? Will it show all in this case as expected?

LATaylor-guardian:
👍

app/controllers/CampaignApi.scala
kelvin-chappell: The idea is `dateRange` will then be `None`, which will mean the filter won't apply in `CampaignReferralRepository`, so then all the campaign figures will be fetched.

kelvin-chappell: So you have to give both dates if you want filtered results, which seems like a reasonable first iteration.

PR #234, opened 2017-11-23
Share counts

rich-nguyen:
Fetches the share counts for a given campaign using Reqwest, and updates the 'latest analytics' state object.

screen shot 2017-11-23 at 10 56 24


public/services/CampaignsApi.js
```diff
@@ -38,11 +39,51 @@ export function fetchLatestAnalytics() {
...
+    return new Promise(function(resolve) {
+      Reqwest(req)
+        .then(res => {
+          values.linkedInShares = res.count;
+          resolve();
+        })
+        .fail(resolve);
+    });
+  });
+  return Promise.all(linkedInShares)
```
kelvin-chappell, line 76: Wouldn't it be better to fetch the share counts in a batch job and cache them in dynamo?

rich-nguyen: I'm not that worried about performance, and the information is coming from an API we have direct access to. It seems simple to just access it?

kelvin-chappell: I thought there was a requirement to report on shares by week etc but if not it's not a problem.  Could be done later anyway if needed.

kelvin-chappell (approved):
👍

public/services/CampaignsApi.js
LATaylor-guardian: @rich-nguyen @kelvin-chappell Maybe worth checking with Chloe or Hannah.

PR #236, opened 2017-11-24
Hide global shares on territory views

rich-nguyen:
Share counts are global, so we need to hide them on uk, us, au territories.


janua:
:+1:

PR #238, opened 2017-11-27
Add dwell time by device and country to performance breakdown.

LATaylor-guardian:


janua:
:+1:

JonNorman (approved):
Looks good to me. How come you're removing the `dataUnavailable` in preference for `0`?

LATaylor-guardian:
> Looks good to me. How come you're removing the dataUnavailable in preference for 0?

All the data is available now (before it wasn't provided by the labs reporting job). So if there is no data available for a particular device then it should show 0.

PR #240, opened 2017-11-30
Add date-range picker for referrals

kelvin-chappell:
image

### DIFF
--- a/public/components/CampaignAnalytics/Analytics/CampaignReferrals.js
+++ b/public/components/CampaignAnalytics/Analytics/CampaignReferrals.js
@@ -5,16 +5,27 @@ import {getCampaignReferrals, setToggleNode} from "../../../actions/CampaignActi
 import CampaignReferralHeader from "./CampaignReferralHeader";
 import InfinityMenu from "react-infinity-menu";
 import ProgressSpinner from "../../utils/ProgressSpinner";
+import DateRangeEditor from "./DateRangeEditor";
 
 class CampaignReferrals extends React.Component {
 
-  componentWillMount() {
-    this.props.campaignReferralActions.getCampaignReferrals(this.props.campaign.id);
+  constructor(props) {
+    super(props);
+    this.state = {
+      dateRange: {
+        startDate: DateRangeEditor.toDate(this.props.campaign.startDate),
+        endDate: DateRangeEditor.toDate(this.props.campaign.endDate)
+      }
+    };
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.campaign.id !== this.props.campaign.id) {
-      this.props.campaignReferralActions.getCampaignReferrals(nextProps.campaign.id);
+  componentWillMount() {
+    this.props.campaignReferralActions.getCampaignReferrals(this.props.campaign.id, this.state.dateRange);
+  }
+
+  componentWillUpdate(nextProps, nextState) {
+    if (nextProps.campaign.id !== this.props.campaign.id || nextState !== this.state) {
+      this.props.campaignReferralActions.getCampaignReferrals(nextProps.campaign.id, nextState.dateRange);
     }
   }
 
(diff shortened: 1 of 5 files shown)


public/components/CampaignAnalytics/Analytics/CampaignReferrals.js
```diff
@@ -5,60 +5,77 @@ import {getCampaignReferrals, setToggleNode} from "../../../actions/CampaignActi
...
   componentWillMount() {
-    this.props.campaignReferralActions.getCampaignReferrals(this.props.campaign.id);
+    this.props.campaignReferralActions.getCampaignReferrals(this.props.campaign.id, this.state.dateRange);
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.campaign.id !== this.props.campaign.id) {
-      this.props.campaignReferralActions.getCampaignReferrals(nextProps.campaign.id);
+  componentWillUpdate(nextProps, nextState) {
+    if (nextProps.campaign.id !== this.props.campaign.id || nextState !== this.state) {
```
rich-nguyen: I'm not sure if that test really captures specific-case update criteria. Is this supposed to detect changes in `dateRange`? could it be more specific? also, the params could just be an object, that encapsulates all the state that parameterises the query (the campaign id, date range). It's also a bit confusing having redux and `componentWillUpdate` overlapping each other.

public/components/CampaignAnalytics/Analytics/DateRangeEditor.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  handleChange({target}) {
+    this.setState({
+      [target.name]: target.value
+    });
+  }
+
+  static zeroPad(num) {
+    return ('0' + num).slice(-2)
```
rich-nguyen, line 27: I recognise this time-formatting snippet from stackoverflow. a classic!

rich-nguyen:
It looks great, just that one issue about the state. Admittedly, it was mainly me moving the filtering out of the component and into the global redux store that has complicated things. But if the three parameters that influenced the tree were in one place, it would be fairly readable (three parameters being date, campaign id, ordering).

public/components/CampaignAnalytics/Analytics/DateRangeEditor.js
kelvin-chappell: The source of all my knowledge.  😄

public/components/CampaignAnalytics/Analytics/CampaignReferrals.js
kelvin-chappell: @rich-nguyen Is this clearer?  I don't think the change in properties is actually relevant so it's just a particular change in state that I need to look out for.

public/components/CampaignAnalytics/Analytics/CampaignReferrals.js
```diff
@@ -5,60 +5,77 @@ import {getCampaignReferrals, setToggleNode} from "../../../actions/CampaignActi
...
   componentWillMount() {
-    this.props.campaignReferralActions.getCampaignReferrals(this.props.campaign.id);
+    this.props.campaignReferralActions.getCampaignReferrals(this.props.campaign.id, this.state.dateRange);
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.campaign.id !== this.props.campaign.id) {
-      this.props.campaignReferralActions.getCampaignReferrals(nextProps.campaign.id);
+  componentWillUpdate(nextProps, nextState) {
+    if (nextState.dateRange !== this.state.dateRange) {
```
rich-nguyen, line 27: what happened to the campaign id test?

rich-nguyen:
👍

public/components/CampaignAnalytics/Analytics/CampaignReferrals.js
kelvin-chappell: I don't think we need it, but will reinstate if anything strange happens.

PR #244, opened 2017-12-04
Permit query operation in Dynamo.

LATaylor-guardian:
This was preventing the media events job to run within the data lake. @rich-nguyen FYI.


rich-nguyen:
Thanks @LATaylor-guardian , I'm not sure how this could ever have worked, now.

PR #246, opened 2017-12-04
Add table for social referrals

kelvin-chappell:
Need a separate table for social referrals, as data has quite different structure.


JonNorman (approved):

LATaylor-guardian:
👍

PR #249, opened 2017-12-12
Show uniques graph contextual to territory selection.

LATaylor-guardian:
- Provides functionality to allow the user to set a target per territory. 
- Makes any target that is being displayed to the user contextual to the territory. If no target is available it will be hidden.


app/controllers/CampaignApi.scala
```diff
@@ -87,17 +87,22 @@ class CampaignApi(components: ControllerComponents, authAction: AuthAction[AnyCo
     result getOrElse InternalServerError
   }
 
-  def getCampaignPageViewsFromDatalake(id: String) = authAction { _ =>
-    CampaignService.getPageViews(id) match {
+  def getCampaignPageViewsFromDatalake(id: String) = authAction { implicit request =>
+    val territory = Territory(request.getQueryString("territory") getOrElse "global")
```
kelvin-chappell: Could add this as an optional param to the endpoint

app/controllers/CampaignApi.scala
```diff
@@ -87,17 +87,22 @@ class CampaignApi(components: ControllerComponents, authAction: AuthAction[AnyCo
...
       case Left(_)                       => InternalServerError
       case Right(pageViews)              => Ok(Json.toJson(pageViews))
     }
   }
 
   // TODO - fix return types
-  def getCampaignUniquesFromDatalake(id: String) = authAction { _ =>
-    CampaignService.getUniquesDataForGraph(id).map(uniquesData => Ok(Json.toJson(uniquesData))) getOrElse NotFound
+  def getCampaignUniquesFromDatalake(id: String) = authAction { implicit request =>
+    val territory = Territory(request.getQueryString("territory") getOrElse "global")
```
kelvin-chappell: same here

app/model/Campaign.scala
```diff
@@ -18,7 +18,7 @@ case class Campaign(
   startDate: Option[DateTime] = None,
   endDate: DateTime,
   targets: Map[String, Long] = Map.empty,
-  campaignTargets: Option[Map[String, Map[String, Long]]] = None,
+  campaignTargets: Option[Map[String, Map[String, Long]]] = Some(Map.empty),
```
kelvin-chappell, line 21: If this is never none, should it be optional?

kelvin-chappell (approved):
👍

app/model/Campaign.scala
LATaylor-guardian: Not all campaigns have this field at the moment which causes it to break.

app/controllers/CampaignApi.scala
LATaylor-guardian: Yes - will change.

PR #250, opened 2017-12-13
Add social referrals to UI

kelvin-chappell:
Referrals component now looks like this: 
image

On-platform and social referrals both respond to the date-range picker.

I haven't enabled the territory choice for social referrals yet, as on-platform referrals don't have any territory data yet so will do both at the same time.  This means that when any other territory is chosen apart from *global* the entire referrals component will disappear.

I'm planning to move the referrals component up to the performance tab further up the page in another PR.


app/model/SocialPlatform.scala
```diff
@@ -12,12 +12,22 @@ case object Facebook extends SocialPlatform {
...
   implicit lazy val writes: Writes[SocialPlatform] = platform => JsString(platform.name)
 
-  def fromDomain(domain: String): SocialPlatform = {
-    domain match {
-      case Facebook.domain => Facebook
-    }
+  def fromDomain(domain: String): SocialPlatform = domain match {
+    case Facebook.domain => Facebook
+    case Twitter.domain  => Twitter
+    case _               => Unknown
```
janua, line 31: I know you probably wouldn't use it, but would it be worth capturing the unknown value? Maybe to reveal what they were later?

kelvin-chappell: It should come out in the UI so should be obvious.  I can't imagine how this case would happen though because each platform will be brought in one by one so should all be known.

PR #255, opened 2017-12-18
Fix facebook share fetching

kelvin-chappell:
By making same call as frontend.

See:
https://github.com/guardian/frontend/blob/master/applications/app/services/FacebookGraphApi.scala#L48


shtukas (approved):

shtukas:
Awesome. Thanks !

PR #257, opened 2018-01-03
Remove social referrals warning

kelvin-chappell:
Have backfilled data from 1 September now so no need for warning about missing data.

This PR also modifies the date-range picker so that it's only possible to choose dates between 1 September and the current date.  Because that's all the data that's available.


LATaylor-guardian:
👍

PR #259, opened 2018-01-10
Use queryEndDate for the uniques graph

janua:
This switches over the uniques graph in the campaign view to use `queryEndDate` instead of `reportExecutionTime` for each graph entry; when we backfill, we will get lots of the same date in `reportExecutionTimestamp` for different dates of the campaign as the backfill will run on the same day for each day.

#### Fixed


kelvin-chappell (approved):
👍

LATaylor-guardian:
Apart from the build failing this looks good. Thanks for fixing it 👍

PR #260, opened 2018-01-11
Upgrade react

janua:
It seemed that some of the react versions that we were pulling in were missing peer dependencies for other libraries. (I think it was react `15.4.0+`).

This upgrades the dependencies, and CircleCI now builds correctly.


kelvin-chappell (approved):
Makes sense 👍

PR #261, opened 2018-01-15
Fix campaign central

janua:
This reverts #260 and also freezes the version to specific version; there is a dependency between the versions that is broken in a minor version upgrade.


LATaylor-guardian:
👍

PR #262, opened 2018-01-19
Dependency Upgrade

janua:
This upgrades Campaign Central to use React 16.x.x; this means that other dependencies also have to upgrade.

`react-router` has caused the most problems; version 4 has some massive changes to how things work. It seems they are moving in a direction to allow the user of the library more choice as to how things are handled, this means that some of the magic gets removed.
The two main things from `react-router` were;
 - Switching from `Link` to `NavLink` as `activeClassName` is no longer supported by `Link`.
 - On from this, the way query parameters are handled has been left to the user, this means that we now have to use a function that compares the query parameters instead of the library magically resolving this for us
 - Some things have moved to `react-router-dom`; mainly anything to do with the browser
 - `react-router` now does not allow nested routes, now you must nest routes within components instead

Also; `prop-types` is a library that contains what used to exist in `react`.


public/components/Sidebar/Sidebar.js
```diff
@@ -14,11 +14,46 @@ class Sidebar extends React.Component {
...
+
+    for (let param in searchValues) {
+      if (searchValues.hasOwnProperty(param)) {
+        if (searchValues[param]) {
+          searchParams.set(param, searchValues[param]);
+        } else {
+          searchParams.delete(param);
+        }
+      }
+    }
```
kelvin-chappell, line 48: Wow!

kelvin-chappell (approved):
Painful work! 👍

LATaylor-guardian:
👍

PR #263, opened 2018-01-22
Fix links and benchmarks

janua:
This brings two fixes;

Once moved over to using `NavList` for the side links, they weren't updating based on what was in the browsers location; this adds `withRouter` on the `connect` of `Sidebar` in order for it to notice location updates.

Also, one of the comparisons in `Benchmarks` was ALWAYS true as the object is deep which would either mean we would have to use a deep comparison; this switched the comparison to use `lodash.isEqual`. (The result of this was swallowing up all of DyanmoDBs provisioned read throughput)


kelvin-chappell (approved):
👍

LATaylor-guardian:
👍 (nice spot)

PR #264, opened 2018-02-06
Capi preview IAM auth

tomrf1:
We are in the process of changing how we authorise access to CAPI preview. In the past we have used basic auth (in Concierge), with the usernames/passwords maintained by the CAPI team.
Instead we will use api-gateway to move authorisation out of concierge and avoid having to maintain usernames/passwords.

The api-gateway uses IAM authorisation. This means the app assumes a cross-account role, and signs each (preview) request using this role.

Note - this means you will now need capi credentials to run the app locally.
I've updated DEV and PROD config files in s3, but left the old preview config items there for now in case we want to roll back.

Tested locally using `curl -XPOST localhost:2267/management/api/refreshExpiringCampaigns`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ You will then need to create a local configuration file. The easiest way to do t
 cp conf/example-local.conf conf/local.conf
 ```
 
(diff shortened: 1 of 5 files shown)


LATaylor-guardian:
👍

